PR #21, opened 2021-10-22
Allow family to use constructors that do not coerce to function pointers

divagant-martian:
Right now a Family can only use as constructors types that coerce to function pointers, but it can be useful to provide more general constructors.

An example is in https://github.com/libp2p/rust-libp2p/pull/2235 where creating an histogram for the score has no clear bucket bounds since those can be affected by the application score, or simply put.. someone wanting to get score metrics might be interested in some custom bounds/buckets. In order to make this parameterizable we would do something like:

` let my_metric = Family::new_with_constructor(move || Histogram::new(exponential_buckets(custom_start, 1.0, 10)));` 
This does not coerce to a function pointer and seems like a reasonable thing to expect to be able to do.

This makes it possible in a way slightly more convoluted but still convenient (I think) without creating a breaking change (See tests for an example).

If you think there is a better way to do it let me know

### DIFF
--- a/src/metrics/family.rs
+++ b/src/metrics/family.rs
@@ -125,7 +125,7 @@ pub struct Family<S, M, C = fn() -> M> {
 /// }
 ///
 /// impl MetricConstructor<Histogram> for CustomBuilder {
-///     fn new(&self) -> Histogram {
+///     fn new_metric(&self) -> Histogram {
 ///         // When a new histogram is created, this function will be called.
 ///         Histogram::new(self.buckets.iter().cloned())
 ///     }
@@ -135,7 +135,7 @@ pub struct Family<S, M, C = fn() -> M> {
 /// let metric = Family::<(), Histogram, CustomBuilder>::new_with_constructor(custom_builder);
 /// ```
 pub trait MetricConstructor<M> {
-    fn new(&self) -> M;
+    fn new_metric(&self) -> M;
 }
 
 /// In cases in which the explicit type of the metric is not required, it is
@@ -151,7 +151,7 @@ pub trait MetricConstructor<M> {
 /// # metric.get_or_create(&());
 /// ```
 impl<M, F: Fn() -> M> MetricConstructor<M> for F {
-    fn new(&self) -> M {
+    fn new_metric(&self) -> M {
         self()
     }
 }
@@ -222,7 +222,7 @@ impl<S: Clone + std::hash::Hash + Eq, M, C: MetricConstructor<M>> Family<S, M, C
         }
 
         let mut write_guard = self.metrics.write().unwrap();
-        write_guard.insert(label_set.clone(), self.constructor.new());
+        write_guard.insert(label_set.clone(), self.constructor.new_metric());
 
         drop(write_guard);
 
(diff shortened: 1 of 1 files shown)


src/metrics/family.rs
```diff
@@ -107,7 +107,17 @@ pub struct Family<S, M> {
     /// [`Histogram`](crate::metrics::histogram::Histogram) in order to set
     /// specific buckets, a custom constructor is set via
     /// [`Family::new_with_constructor`].
-    constructor: fn() -> M,
+    constructor: C,
+}
+
+pub trait MetricConstructor<M> {
```
mxinden, line 113: Would you mind adding some documentation pointing to `Family::new_with_constructor` and a doc example to this type?

src/metrics/family.rs
```diff
@@ -97,7 +97,7 @@ use std::sync::{Arc, RwLock, RwLockReadGuard};
 /// # assert_eq!(expected, String::from_utf8(buffer).unwrap());
 /// ```
 // TODO: Consider exposing hash algorithm.
-pub struct Family<S, M> {
+pub struct Family<S, M, C: MetricConstructor<M> = fn() -> M> {
```
mxinden, line 100:
```suggestion
pub struct Family<S, M, C = fn() -> M> {
```
Is this bound needed?

src/metrics/family.rs
```diff
@@ -119,7 +129,7 @@ impl<S: Clone + std::hash::Hash + Eq, M: Default> Default for Family<S, M> {
     }
 }
 
-impl<S: Clone + std::hash::Hash + Eq, M> Family<S, M> {
+impl<S: Clone + std::hash::Hash + Eq, M, C: MetricConstructor<M>> Family<S, M, C> {
```
mxinden, line 132:
```suggestion
impl<S: Clone + std::hash::Hash + Eq, M, C> Family<S, M, C> {
```

Is this bound needed?

src/metrics/family.rs
```diff
@@ -192,16 +202,16 @@ impl<S: Clone + std::hash::Hash + Eq, M> Family<S, M> {
     }
 }
 
-impl<S, M> Clone for Family<S, M> {
+impl<S, M, C: MetricConstructor<M> + Clone> Clone for Family<S, M, C> {
```
mxinden, line 205:
```suggestion
impl<S, M, C: Clone> Clone for Family<S, M, C> {
```

Is this bound needed?

src/metrics/family.rs
```diff
@@ -192,16 +202,16 @@ impl<S: Clone + std::hash::Hash + Eq, M> Family<S, M> {
...
         Family {
             metrics: self.metrics.clone(),
-            constructor: self.constructor,
+            constructor: self.constructor.clone(),
         }
     }
 }
 
-impl<S, M: TypedMetric> TypedMetric for Family<S, M> {
+impl<S, M: TypedMetric, C: MetricConstructor<M>> TypedMetric for Family<S, M, C> {
```
mxinden, line 214:
```suggestion
impl<S, M: TypedMetric, C> TypedMetric for Family<S, M, C> {
```

Is this bound needed?

mxinden (commented):
Sorry for the delay here @divagant-martian!

I think this is a valid approach and not too intrusive.

src/metrics/family.rs
divagant-martian: I added docs and an example to the trait, and also to the `Fn() -> M` implementor for completeness. Thanks for the suggestion

mxinden (approved):
Thank you for the elaborate work!

PR #24, opened 2021-11-09
*: Prepare v0.12.1

mxinden:


mxinden:
[Tagged](https://github.com/mxinden/rust-open-metrics-client/releases/tag/v0.12.1) and [published](https://crates.io/crates/open-metrics-client/0.12.1).

//CC @divagant-martian

mxinden:
I yanked the release due to https://github.com/libp2p/rust-libp2p/issues/2333#issuecomment-964668411. I will most likely cut a new release (`v0.13.0`) tomorrow.

divagant-martian:
I'm surprised about this error since there is a test with that specific function. I'll look into it

divagant-martian:
So it looks like the type inference breaks. The failing lines are fixed by specifying the types. I don't see an alternative and thus would make it a breaking change indeed

PR #35, opened 2022-01-19
src/lib: Make imports visible in example code

### LINKED ISSUE #34: Hidden imports in lib.rs example code
Hello folks,

Looking at the example code in [lib.rs](https://github.com/prometheus/client_rust/blob/master/src/lib.rs) (when rendered on docs.rs), it was a little confusing to figure out what was going on due to many functions and types from the client_rust library being imported but the imports themselves being hidden.

What do you think about un-hiding the library imports so that it's more clear which parts of the client_rust library the example code is referencing?

56quarters:
Fixes #34

Signed-off-by: Nick Pillitteri <nick.pillitteri@grafana.com>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,13 +11,13 @@
 //! # Examples
 //!
 //! ```
-//! # use prometheus_client::encoding::text::Encode;
-//! # use prometheus_client::encoding::text::encode;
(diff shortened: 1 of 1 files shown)


mxinden (approved):
That was fast. Thanks!

Issue #36, opened 2022-01-20
Const Metrics


phyber:
Hi,

Thanks for creating an official Rust crate for Prometheus. :)

I'm wondering how we should handle [Const Metrics](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus?utm_source=godoc#NewConstMetric) with this crate. Are these supported at the moment? If they are, it's a little non-obvious.

My use case is for exporting counters from the operating system. The OS itself guarantees that these are always increasing (unless the counter wraps, etc), but they're difficult to export with only the `inc` and `inc_by` methods as additional variables must be maintained to work out the difference between the old OS counter and the current OS counter, so we can finally `inc_by`.

If Const Metrics are not currently supported, please consider this a feature request.

Thanks!

mxinden:
> Thanks for creating an official Rust crate for Prometheus. :)

Happy that it is of some use :heart: 

> My use case is for exporting counters from the operating system. The OS itself guarantees that these are always increasing (unless the counter wraps, etc), but they're difficult to export with only the `inc` and `inc_by` methods as additional variables must be maintained to work out the difference between the old OS counter and the current OS counter, so we can finally `inc_by`.

In this particular use-case, would a metric type called _constant_ not be misleading? In other words, the use-case here produces a non-constant metric, thus one would not expect to be using a metric type with _const_ in its name, no?

The crate allows you to access the inner (atomic) type of a `Counter` via [`Counter::inner`](https://docs.rs/prometheus-client/0.15.0/prometheus_client/metrics/counter/struct.Counter.html#method.inner) which then allows you to set the value of the counter. The API contract though is, that you uphold the guarantees of a Prometheus _Counter_:

``` rust
    /// Exposes the inner atomic type of the [`Counter`].
    ///
    /// This should only be used for advanced use-cases which are not directly
    /// supported by the library.
    ///
    /// The caller of this function has to uphold the property of an Open
    /// Metrics counter namely that the value is monotonically increasing, i.e.
    /// either stays the same or increases.
    pub fn inner(&self) -> &A {
```
With this little trick in mind, would the following work for you?

``` rust
let counter: Counter = Counter::default();

// Register the counter with a registry.

// Update the counter value to represent the OS value.
counter.inner().store(os_value);
```

In addition to achieving your goal, your metric would also properly be tagged with the OpenMetrics _Counter_ type.

> If they are, it's a little non-obvious.

[...]

PR #41, opened 2022-01-24
src/encoding/text.rs: Expose Encoder methods

mxinden:
By exposing the various `Encoder` builder methods, downstream users can
implement their custom metrics using `EncodeMetric`.

See https://github.com/prometheus/client_rust/issues/36 for additional context.

//CC @phyber


phyber:
Sorry, I forgot about this. I'll give it a look soon :)

phyber:
This seems perfect, I was able to quickly get my test working with this (which was almost 1:1 with the example you added) :)

phyber (approved):

PR #43, opened 2022-02-04
Esp32 support

sh0rez:
This library already works really well on the embedded [ESP32](http://esp32.net/) SOC, but that platform being 32-bit means we cannot use `AtomicU32`.

This PR adds `xtensa` to the list of 32-bit architectures, so that it compiles.
It also implements `Atomic` for `f32`, so that Gauges become usable.

IMHO we really should use https://github.com/rust-lang/rust/issues/32976 once available, but that is for another day.


mxinden (commented):
Thanks for providing a patch!

Would you mind adding a changelog entry? Something along the lines of:

```
## [0.15.2] - unreleased

### Added

- Support for `f32` `Gauge` and `Counter`. See [PR xx].
- Support for `xtensa` architecture. See [PR xx].

```

And bump the patch version in `Cargo.toml`?

Also does Rust support cross compilation for `xtensa`? If so, want to add it to the matrix in our CI config?

PR #47, opened 2022-02-24
Protobuf support

### LINKED ISSUE #46: Custom encoder?
I would like to send (partial) metrics stored the `Registry` in original JSON format differ from OpenMetrics, into an endpoint different from Prometheus server. 

Here is a pseudo code, shows what I want to do:

```rust

        // (omitted)

        // Using the OpenMetrics format.
        let mut buffer = vec![];
        encode(&mut buffer, &registry).unwrap();

        // On the other hand, send (partial) metrics in JSON, into an endpoint different from Prometheus server.
        let mut data = vec![];
        for (descriptor, metric) in registry.iter() {
            if descriptor.name() == "__test__" {
                data.push(
                    translate_to_json(metric), // original JSON format
                );
            }
        }
        send(data); // send to an endpoint different from Prometheus server
```

I think I need to implement custom encoder, translate metrics to JSON format.

Please let me know if there is any good way. 🙏

ackintosh:
Closes #46 

This PR adds protobuf support to the client, which provides functionality to encode metrics to an instance of the [OpenMetrics protobuf](https://github.com/OpenObservability/OpenMetrics/blob/main/proto/openmetrics_data_model.proto) format.

```rust

        // ...

        // Encode metrics to an instance of the OpenMetrics protobuf format.
        let metric_set: openmetrics_data_model::MetricSet = encode(&registry);

        for family in metric_set.metric_families {
            println!(family.name);

            for metric in family.metrics {
                // ...
            }
        }
```

## 📝 Todos

- [x] Implement encoding
- [x] Implement unit tests
- [x] Documentation
- [x] Fix how to manage the dependencies of `.proto` files (noted below)
- [x] Implement procedural macro with reference to [`derive-text-encode`](https://github.com/prometheus/client_rust/blob/66be8e2a1bfcdd13a6d8f1fde7fd42025d388f2f/derive-text-encode/src/lib.rs#L9)
- [x] Benchmark code
- [ ] ~~Integration test on lighthouse~~
  - I found out that we need another enhancement for this client in order to test on lighthouse. (https://github.com/sigp/lighthouse/issues/2956#issuecomment-1086415339)
  - So I wouldn't (can't) test on lighthouse at this time.

FYI: I'm working to use this Protobuf support, in sigp/gossipsub-testground#4, in order to record metrics whose gossipsub-libp2p to InfluxDB.

--- 

## (Edit: Resolved) How to manage the dependencies of `.proto` files

This PR introduces a `openmetrics_data_model.proto` file dependency. We can get the file from [here](https://github.com/OpenObservability/OpenMetrics/blob/main/proto/openmetrics_data_model.proto).

For now, I've copied the .proto file manually and added it to git. However, I think that is not good as it could happen operational errors. So we need to consider how to manage the .proto file.

1. Using [Buf](https://github.com/bufbuild/buf)

- Doc: https://docs.buf.build/tour/add-a-dependency#depend-on-googleapis
- This looks good, I've just read the doc though.
- However, in order to use Buf as dependency management, **`openmetrics_data_model.proto` file should be pushed to [BSR](https://docs.buf.build/bsr/overview), which is a proto file registry**.
  - There's not the proto file in BSR: https://buf.build/OpenObservability 🙅‍♂️ 

2. Download the proto file via `build.rs`

- Instead of using something tools like Buf, we can download the proto file via `build.rs`.
- For example: https://github.com/sigp/lighthouse/blob/stable/common/deposit_contract/build.rs#L93-L95
  - download the file from the GitHub repository, with a specific commit hash
  - verify the checksum of the file downloaded

~~I couldn't find another way, **I think `2. Download the proto file via build.rs` is good**.~~

Edit: I changed my mind, manually coping is suitable: https://github.com/prometheus/client_rust/pull/47#issuecomment-1100676043

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ prost-types = "0.9.0"
 async-std = { version = "1", features = ["attributes"] }
 criterion = "0.3"
 http-types = "2"
-pyo3 = "0.15"
+pyo3 = "0.16"
 quickcheck = "1"
 rand = "0.8.4"
 tide = "0.16"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus-client"
-version = "0.18.0"
+version = "0.18.1"
 authors = ["Max Inden <mail@max-inden.de>"]
 edition = "2021"
 description = "Open Metrics client library allowing users to natively instrument applications."
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ pyo3 = "0.16"
 quickcheck = "1"
 rand = "0.8.4"
 tide = "0.16"
+actix-web = "4"
 
 [build-dependencies]
 prost-build = { version = "0.9.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ homepage = "https://github.com/prometheus/client_rust"
 documentation = "https://docs.rs/prometheus-client"
 
 [features]
-protobuf = ["dep:prost", "dep:prost-types", "dep:prost-build", "dep:prometheus-client-derive-proto-encode"]
+protobuf = ["dep:prost", "dep:prost-types", "dep:prost-build", "dep:prometheus-client-derive-proto-encode", "dep:void"]
 
 [workspace]
 members = ["derive-text-encode", "derive-proto-encode"]
@@ -24,7 +24,7 @@ prometheus-client-derive-proto-encode = { version = "0.1.0", path = "derive-prot
 prometheus-client-derive-text-encode = { version = "0.3.0", path = "derive-text-encode" }
 prost = { version = "0.9.0", optional = true }
 prost-types = { version = "0.9.0", optional = true }
-void = "1.0"
+void = { version = "1.0", optional = true }
 
 [dev-dependencies]
 async-std = { version = "1", features = ["attributes"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,15 +11,16 @@ homepage = "https://github.com/prometheus/client_rust"
 documentation = "https://docs.rs/prometheus-client"
 
 [features]
-protobuf = ["dep:prost", "dep:prost-types", "dep:prost-build"]
+protobuf = ["dep:prost", "dep:prost-types", "dep:prost-build", "dep:prometheus-client-derive-proto-encode"]
 
 [workspace]
-members = ["derive-text-encode"]
+members = ["derive-text-encode", "derive-proto-encode"]
 
 [dependencies]
 dtoa = "1.0"
 itoa = "1.0"
 owning_ref = "0.4"
+prometheus-client-derive-proto-encode = { version = "0.1.0", path = "derive-proto-encode", optional = true }
 prometheus-client-derive-text-encode = { version = "0.3.0", path = "derive-text-encode" }
 prost = { version = "0.9.0", optional = true }
 prost-types = { version = "0.9.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus-client"
-version = "0.15.1"
+version = "0.16.0"
 authors = ["Max Inden <mail@max-inden.de>"]
 edition = "2018"
 description = "Open Metrics client library allowing users to natively instrument applications."
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus-client"
-version = "0.17.0"
+version = "0.18.0"
 authors = ["Max Inden <mail@max-inden.de>"]
 edition = "2021"
 description = "Open Metrics client library allowing users to natively instrument applications."
@@ -19,7 +19,7 @@ members = ["derive-text-encode", "derive-proto-encode"]
 [dependencies]
 dtoa = "1.0"
 itoa = "1.0"
-owning_ref = "0.4"
+parking_lot = "0.12"
 prometheus-client-derive-proto-encode = { version = "0.1.0", path = "derive-proto-encode", optional = true }
 prometheus-client-derive-text-encode = { version = "0.3.0", path = "derive-text-encode" }
 prost = { version = "0.9.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,3 +47,8 @@ harness = false
 name = "text"
 path = "benches/encoding/text.rs"
 harness = false
+
+[[bench]]
+name = "proto"
+path = "benches/encoding/proto.rs"
+harness = false
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,8 @@
 [package]
 name = "prometheus-client"
-version = "0.16.0"
+version = "0.17.0"
 authors = ["Max Inden <mail@max-inden.de>"]
-edition = "2018"
+edition = "2021"
 description = "Open Metrics client library allowing users to natively instrument applications."
 license = "Apache-2.0 OR MIT"
 keywords = ["openmetrics", "prometheus", "metrics", "instrumentation", "monitoring"]
@@ -17,7 +17,7 @@ members = ["derive-text-encode"]
 dtoa = "1.0"
 itoa = "1.0"
 owning_ref = "0.4"
-prometheus-client-derive-text-encode = { version = "0.2.0", path = "derive-text-encode" }
+prometheus-client-derive-text-encode = { version = "0.3.0", path = "derive-text-encode" }
 prost = "0.9.0"
 prost-types = "0.9.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ prometheus-client-derive-proto-encode = { version = "0.1.0", path = "derive-prot
 prometheus-client-derive-text-encode = { version = "0.3.0", path = "derive-text-encode" }
 prost = { version = "0.9.0", optional = true }
 prost-types = { version = "0.9.0", optional = true }
+void = "1.0"
 
 [dev-dependencies]
 async-std = { version = "1", features = ["attributes"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ repository = "https://github.com/prometheus/client_rust"
 homepage = "https://github.com/prometheus/client_rust"
 documentation = "https://docs.rs/prometheus-client"
 
+[features]
+protobuf = ["dep:prost", "dep:prost-types", "dep:prost-build"]
+
 [workspace]
 members = ["derive-text-encode"]
 
@@ -18,8 +21,8 @@ dtoa = "1.0"
 itoa = "1.0"
 owning_ref = "0.4"
 prometheus-client-derive-text-encode = { version = "0.3.0", path = "derive-text-encode" }
-prost = "0.9.0"
-prost-types = "0.9.0"
+prost = { version = "0.9.0", optional = true }
+prost-types = { version = "0.9.0", optional = true }
 
 [dev-dependencies]
 async-std = { version = "1", features = ["attributes"] }
@@ -31,7 +34,7 @@ rand = "0.8.4"
 tide = "0.16"
 
 [build-dependencies]
-prost-build = "0.9.0"
+prost-build = { version = "0.9.0", optional = true }
 
 [[bench]]
 name = "family"
--- a/src/encoding/proto.rs
+++ b/src/encoding/proto.rs
@@ -272,6 +272,12 @@ impl EncodeGaugeValue for i64 {
     }
 }
 
+impl EncodeGaugeValue for u64 {
+    fn encode(&self) -> openmetrics_data_model::gauge_value::Value {
+        openmetrics_data_model::gauge_value::Value::IntValue(*self as i64)
+    }
+}
+
 impl EncodeGaugeValue for f64 {
     fn encode(&self) -> openmetrics_data_model::gauge_value::Value {
         openmetrics_data_model::gauge_value::Value::DoubleValue(*self)
--- a/src/encoding/proto.rs
+++ b/src/encoding/proto.rs
@@ -834,6 +834,53 @@ mod tests {
     }
 
     #[test]
+    fn encode_family_and_counter_and_histogram() {
+        let mut registry = Registry::<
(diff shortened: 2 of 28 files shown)


ackintosh:
For now we get the compile error:

```bash
error[E0277]: the trait bound `Box<dyn text::SendEncodeMetric>: proto::EncodeMetric` is not satisfied
   --> src/encoding/proto.rs:151:33
    |
151 |         println!("{:?}", encode(&registry));
    |                          ------ ^^^^^^^^^ the trait `proto::EncodeMetric` is not implemented for `Box<dyn text::SendEncodeMetric>`
    |                          |
    |                          required by a bound introduced by this call
    |
    = help: the following implementations were found:
              <Box<(dyn proto::EncodeMetric + 'static)> as proto::EncodeMetric>
              <Box<(dyn proto::SendEncodeMetric + 'static)> as proto::EncodeMetric>
note: required by a bound in `proto::encode`
   --> src/encoding/proto.rs:13:8
    |
11  | pub fn encode<M>(registry: &Registry<M>) -> openmetrics_data_model::MetricSet
    |        ------ required by a bound in this
12  | where
13  |     M: EncodeMetric,
    |        ^^^^^^^^^^^^ required by this bound in `proto::encode`
```

---

> error[E0277]: the trait bound `Box<dyn text::SendEncodeMetric>`

This error message refers to `SendEncodeMetric` trait in the "text", not "proto".

I think, it seems [this definition of Registry](https://github.com/ackintosh/client_rust/blob/66be8e2a1bfcdd13a6d8f1fde7fd42025d388f2f/src/registry.rs#L61) somehow could relates the error. 

I'm not sure how solve the compile error. 🤔 

@mxinden I would appreciate it if you could give me some advice. 🙇

mxinden:
> @mxinden I would appreciate it if you could give me some advice. bow

https://github.com/ackintosh/client_rust/pull/1 should fix your  issue @ackintosh. Let me know if that works for you.

ackintosh:
It works, thank your @mxinden !

mxinden:
Please ping me once this is ready for a first review @ackintosh. Thanks for the help!

mxinden:
This is great news @ackintosh. Unfortunately I will be offline for a bit, thus I will likely not get to it next week :/

src/encoding/proto.rs
```diff
@@ -0,0 +1,839 @@
...
+        }
+    }
+}
+
+/// Trait implemented by each metric type, e.g. [`Counter`], to implement its encoding.
+pub trait EncodeMetric {
+    fn encode(
+        &self,
+        labels: Vec<openmetrics_data_model::Label>,
+    ) -> Vec<openmetrics_data_model::Metric>;
```
mxinden, line 83: Disclaimer: I haven't fully thought this through.

Would it be an option to return an iterator (e.g. `Self::Iterator`) here? That would eliminate the many `vec!` allocations and would only require a single final `collect`.

src/encoding/proto.rs
```diff
@@ -0,0 +1,839 @@
...
+        self.deref().encode(labels)
+    }
+
+    fn metric_type(&self) -> MetricType {
+        self.deref().metric_type()
+    }
+}
+
+pub trait EncodeLabel {
+    fn encode(&self) -> Vec<openmetrics_data_model::Label>;
```
mxinden, line 119: Same here: Would it be an option to return an iterator instead of a vector?

mxinden (commented):
Thanks for the work @ackintosh! I have a couple of high-level comments. I will take a more in-depth look.

mxinden:
> I couldn't find another way, **I think `2. Download the proto file via build.rs` is good**. memo

I would like the `prometheus` crate to be able to build offline, i.e. once one downloaded all dependencies from crates.io to not require internet access. Downloading the file on each build would thus not work.

mxinden:
> Using [Buf](https://github.com/bufbuild/buf)

This is an interesting suggestion. What do you think of moving it to https://github.com/OpenObservability/OpenMetrics/ as it concerns the whole project?

mxinden:
> For now, I've copied the .proto file manually and added into git.

I don't expect any breaking changes to the `.proto` file, and thus think this is a valid option.

Cargo.toml
```diff
@@ -18,6 +18,8 @@ dtoa = "1.0"
 itoa = "1.0"
 owning_ref = "0.4"
 prometheus-client-derive-text-encode = { version = "0.2.0", path = "derive-text-encode" }
+prost = "0.9.0"
```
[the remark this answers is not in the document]
ackintosh: That makes sense, thank you!

src/encoding/proto.rs
ackintosh: Note to self: [Returning Rust Iterators | Depth-First](https://depth-first.com/articles/2020/06/22/returning-rust-iterators/)

ackintosh: Note to self: 

- I tried `Associated Type` which is described [here](https://depth-first.com/articles/2020/06/22/returning-rust-iterators/#associatedtype).
- Commit log: https://github.com/ackintosh/client_rust/commit/e4c46dfcd25eba32801229fb3d43e7fb9ca08338 (on another branch `protobuf-support-associated-type`)
- Build error below occurs:

```text
error[E0207]: the lifetime parameter `'a` is not constrained by the impl trait, self type, or predicates
   --> src/encoding/proto.rs:199:6
    |
199 | impl<'a, N, A> EncodeMetric for Counter<N, A>
    |      ^^ unconstrained lifetime parameter
```

ackintosh: Next, I will try [`Simple Iterator Delegation`](https://depth-first.com/articles/2020/06/22/returning-rust-iterators/#simpleiteratordelegation).

mxinden: @ackintosh thanks for investigating. Let me know in case you want me to take a look as well.

ackintosh: I have fixed the [build error](https://github.com/prometheus/client_rust/pull/47#discussion_r933917283) and now it works well (tests are green).

Here is a PR: https://github.com/ackintosh/client_rust/pull/2

However, the allocation count seems not reduced as far as I checked. 🤔 [Here](https://github.com/ackintosh/client_rust/pull/2#issuecomment-1203634863) is a detail of how I checked the allocation count.

@mxinden Please take a look at https://github.com/ackintosh/client_rust/pull/2 and let me know if you have any concerns. 🙏

ackintosh:
> > Using [Buf](https://github.com/bufbuild/buf)
> 
> This is an interesting suggestion. What do you think of moving it to https://github.com/OpenObservability/OpenMetrics/ as it concerns the whole project?

💡 I posted an issue to OpenMetrics just now: https://github.com/OpenObservability/OpenMetrics/issues/253

ackintosh:
> > I couldn't find another way, **I think `2. Download the proto file via build.rs` is good**. memo
> 
> I would like the `prometheus` crate to be able to build offline, i.e. once one downloaded all dependencies from crates.io to not require internet access. Downloading the file on each build would thus not work.

Ahh, that makes sense. 💡

ackintosh:
FYI: I'm planning that use the Protobuf support, in [gossipsub-testground](https://github.com/sigp/gossipsub-testground/pull/4), in order to record metrics whose gossipsub-libp2p to InfluxDB.

src/encoding/proto.rs
mxinden: @ackintosh is https://github.com/ackintosh/client_rust/pull/5 of any help?

ackintosh: Thank you for the patch @mxinden ! I have learned a lot. ✨ Will merge later.

mxinden: Wonderful. Ping me in case I should do another review or in case you need any help.

ackintosh:
@mxinden 
I have added two tests and some fixes, in anticipation of [use in libp2p-gossipsub](https://github.com/libp2p/rust-libp2p/blob/4db881d95f9bd661d274b734c814ef766a84c400/protocols/gossipsub/src/metrics.rs#L198-L293). In libp2p-gossipsub, Registry treats Family, Counter, and Histogram.

Tests added: 

1. [encode_family_counter_histogram](https://github.com/ackintosh/client_rust/blob/35178c7c281d4812705cf1136178645f791ef586/src/encoding/proto.rs#L804-L834)
1. [encode_family_and_counter_and_histogram](https://github.com/ackintosh/client_rust/blob/35178c7c281d4812705cf1136178645f791ef586/src/encoding/proto.rs#L836-L881)

However, the second test ([encode_family_and_counter_and_histogram](https://github.com/ackintosh/client_rust/blob/35178c7c281d4812705cf1136178645f791ef586/src/encoding/proto.rs#L836-L881)) is failing and I'm not sure how fix it. 🤔 I'm tackling this, I would appreciate it if you could help me with this. 🙏

ackintosh:
Btw, I'm doing a benchmark and comparing iter version and vec version. I'll share you result of it later.

ackintosh:
@mxinden 

> Btw, I'm doing a benchmark and comparing iter version and vec version.

I created a benchmark and test (to count allocations) on my personal sandbox repo. We can switch the iter version and vec version to test, [some changes](https://github.com/ackintosh/sandbox/blob/master/rust/crate-critcmp/src/lib.rs#L2-L9) are required though.

- [Benchmark code](https://github.com/ackintosh/sandbox/blob/master/rust/crate-critcmp/benches/critcmp.rs#L22-L29)
- [Test code to count allocations](https://github.com/ackintosh/sandbox/blob/master/rust/crate-critcmp/src/lib.rs#L127-L137)

And compared the results between iter version and vec version:

#### Benchmark

The iterator version (`baseline-iter`) is a bit slow.

```
$ critcmp baseline-iter baseline-vec

group     baseline-iter                          baseline-vec
-----     -------------                          ------------
encode    1.02      6.6±0.06ms        ? ?/sec    1.00      6.4±0.24ms        ? ?/sec
```

#### Allocation count

- iter: 124214
- vec: 124114

---

The slowness is related to [this](https://github.com/ackintosh/client_rust/blob/aed6113005870c56835b2521f9c897b4726d08b2/derive-proto-encode/src/lib.rs#L33-L42) proc macro implementation.

As far as I investigated, I think [from_generator](https://doc.rust-lang.org/nightly/std/iter/fn.from_generator.html) may improve that but it is a nightly-only experimental API. 

Please let me know if you have any ideas. 🙏

ackintosh:
Now I'm stucking on these issues: 

1. https://github.com/prometheus/client_rust/pull/47#issuecomment-1213637603
1. https://github.com/prometheus/client_rust/pull/47#issuecomment-1214564100

Hmm... how about reverting to return a vector? @mxinden

mxinden:
I took another look. What do you think of https://github.com/ackintosh/client_rust/pull/6?

- It does not require an associated type.
- It does not force any allocations, not even on the implementation of `Family`.
- It significantly reduces the type signatures, both on the implementor and the consumer side.
- Tests all pass.

What do you think @ackintosh?

(I am sorry for leading you astray suggesting to use an `Iterator`.)

ackintosh:
Thank you @mxinden! It's awesome. ✨ I have merged ackintosh#6 and made some updates according to the changes in ackintosh#6.

I run benchmark again in [the same way](https://github.com/prometheus/client_rust/pull/47#issuecomment-1214564100), and the result was definitely good. 

#### Benchmark

```
$ critcmp baseline-vec baseline-max

group     baseline-max                           baseline-vec
-----     ------------                           ------------
encode    1.00      6.1±0.14ms        ? ?/sec    1.05      6.4±0.24ms        ? ?/sec
```

#### Allocation count

- vec: 124114
- now: 114014

---

No worries, the experiments to use an `Iterator` was so exciting and I have learned a lot from them. 😃

ackintosh:
TODOs on this PR have been done. ✅

Cargo.toml
```diff
@@ -10,14 +10,21 @@ repository = "https://github.com/prometheus/client_rust"
...
+
 [workspace]
-members = ["derive-text-encode"]
+members = ["derive-text-encode", "derive-proto-encode"]
 
 [dependencies]
 dtoa = "1.0"
 itoa = "1.0"
 parking_lot = "0.12"
+prometheus-client-derive-proto-encode = { version = "0.1.0", path = "derive-proto-encode", optional = true }
```
mxinden, line 23: Exploring the thought on whether we could use the same derive macro for both text and protobuf. I guess we can not feature flag the generated protobuf code via `#[cfg(feature = "protobuf")]` as we would not have access to the flag in the downstream user crate?

Cargo.toml
```diff
@@ -10,14 +10,21 @@ repository = "https://github.com/prometheus/client_rust"
 homepage = "https://github.com/prometheus/client_rust"
 documentation = "https://docs.rs/prometheus-client"
 
+[features]
+protobuf = ["dep:prost", "dep:prost-types", "dep:prost-build", "dep:prometheus-client-derive-proto-encode", "dep:void"]
```
mxinden, line 14: I think CI is currently not running with `--all-features`. Can you add it with this pull request to make sure your feature is being tested?

mxinden (commented):
Great progress here.

Would you mind bumping the crate version and adding a changelog entry? As far as I can tell, this does not introduce any breaking changes and could thus be released as `v0.18.1`.

Cargo.toml
ackintosh: That is cool if we use the same derive macro, I'll do some experiments. 🙂

ackintosh: I investigated but yeah, I couldn't find a way to do that.

ackintosh:
https://github.com/prometheus/client_rust/pull/47/checks?check_run_id=8046799148
> DCO
> There are 49 commits incorrectly signed off. This means that the author(s) of these commits failed to include a Signed-off-by line in their commit message.

I need to rebase this branch to add Signed-off-by lines.

Cargo.toml
ackintosh: https://github.com/prometheus/client_rust/pull/47/commits/33c940a6e7f8d95fbf4479c6e2301f40bff4670c

ackintosh:
- Bump the crate version and add a changelog entry https://github.com/prometheus/client_rust/pull/47/commits/250db460857b5ee1baadb4913ae07413968574bb
- Rebased this branch according to the DCO check https://github.com/prometheus/client_rust/pull/47#issuecomment-1229102302

ackintosh:
> Rebased this branch according to the DCO check https://github.com/prometheus/client_rust/pull/47#issuecomment-1229102302

Hmm, I noticed that the commit history was broken due to the rebasing... sorry.

I'll create another pull request and close this. 🙇

ackintosh:
I have created https://github.com/prometheus/client_rust/pull/83

Issue #49, opened 2022-02-28
Custom collector for multiple metrics


sd2k:
Hi! I've been looking at implementing a Prometheus collector for the recently announced [tokio-metrics](https://docs.rs/tokio-metrics/latest/tokio_metrics/index.html) crate. Every scrape, I'd like to gather runtime metrics for the currently Tokio runtime. The problem is that doing so requires a non-trivial amount of up-front work to aggregate all of the stats across the N workers in the runtime, which I'd rather not do during every metric's `encode` function (following the [custom metric example](https://github.com/prometheus/client_rust/blob/master/examples/custom-metric.rs)).

Instead I _think_ it'd be ideal if there was a way to do something similar to the [client_python Custom Collector example](https://github.com/prometheus/client_python#custom-collectors), which allows custom collectors to record values for multiple metrics at each scrape time - that'd avoid me having to duplicate work (non-atomically) on every scrape. Do you think such an API would be possible?

Alternatively if you know of another pattern to get around this, I'd love to hear it!

mxinden:
As you stated above, a clean solution is not possible today, i.e. would require some changes within `prometheus-client`. I am not opposed to supporting this use-case, though I need to put more thoughts into the design of a clean abstraction. Design suggestions are most welcome.

In the meantime would sharing some state between the custom metrics be an option? That shared state would be updated with a new aggregate by a single custom metric per scrape. Consecutive custom metrics within the same scrape don't need to aggregate themselves, but instead access the shared state.

sd2k:
Thanks for the reply! Yep, I thought as much, no worries - I'll have a think about potential APIs.

> In the meantime would sharing some state between the custom metrics be an option? That shared state would be updated with a new aggregate by a single custom metric per scrape. Consecutive custom metrics within the same scrape don't need to aggregate themselves, but instead access the shared state.

Yeah, I've got a solution where each custom metric shares a reference to some shared state, but there's a few things about it that are suboptimal:

1. for the metrics to be `'static` they can't hold references, so I'm having to use an `Rc<RefCell<State>>` instead (this isn't too bad though)
2. the only way I can determine whether one metric's `encode` call is within the same scrape as another metric's `encode` call is by either:
   - storing the `Instant` that the state was last updated (and only updating the state if it's more than some `Duration` ago), _or_
[...]

PR #54, opened 2022-03-18
Add example for actix-web framework

### LINKED ISSUE #53: Add example for actix-web framework
Hi guys, thanks for your contribution!

There is an issue with using the client with actix-web framework.
Because `Registry` doesn't implement `Copy` - the only working solution I've found is to wrap it with `Mutex`. But it requires locking at least on each collecting of metrics which isn't good for performance. Are there any other suggestions?

oriontvv:
Resolves #53


examples/actix-web.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+#[derive(Clone, Hash, PartialEq, Eq, Encode)]
+pub struct RequestTypeLabels {
+    pub request_type: RequestType,
+}
+
+/// Holds all metrics.
+/// We shouldn't store metrics inside a MetricsCollector -
+/// overwise we would have to take a lock for each increment of any metric
```
mxinden, line 22:
```suggestion
/// otherwise we would have to take a lock for each increment of any metric.
```

examples/actix-web.rs
```diff
@@ -0,0 +1,103 @@
+use std::sync::Mutex;
+
+use actix_web::{web, App, HttpResponse, HttpServer, Responder, Result};
+use prometheus_client::encoding::text::{encode, Encode};
+use prometheus_client::metrics::counter::Counter;
+use prometheus_client::metrics::family::Family;
+use prometheus_client::registry::Registry;
+
+#[derive(Clone, Hash, PartialEq, Eq, Encode)]
+pub enum RequestType {
```
mxinden, line 10:
```suggestion
pub enum Method {
```

In the `tide` example we call it `Method`. I don't really care about the names, though I would prefer to be consistent. What do you think?

https://github.com/prometheus/client_rust/blob/master/examples/tide.rs

examples/actix-web.rs
```diff
@@ -0,0 +1,103 @@
...
+pub async fn some_handler(metrics: web::Data<Metrics>) -> impl Responder {
+    metrics.inc_requests(RequestType::Get);
+    format!("okay")
+}
+
+#[actix_web::main]
+async fn main() -> std::io::Result<()> {
+    let metrics = web::Data::new(Metrics::new());
+    // We have to wrap it with Mutex because `Registry` doesn't implement Copy-trait
+    let metrics_collector = web::Data::new(Mutex::new(MetricsCollector::new(&metrics)));
```
mxinden, line 91: Note that `encode` takes an immutable reference to `Registry`. Thus there is no need for a `Mutex` in case you never register new metrics after the first Prometheus scrape. An `Arc` should suffice.

examples/actix-web.rs
```diff
@@ -0,0 +1,103 @@
...
+        self.requests
+            .get_or_create(&RequestTypeLabels {
+                request_type: request_type,
+            })
+            .inc();
+    }
+}
+
+/// Registers and collects metrics
+pub struct MetricsCollector {
```
mxinden, line 44: Why this wrapper around `Registry`? Despite the fact that `client_rust` does use the `Collector` abstraction (yet see https://github.com/prometheus/client_rust/issues/49), if I am not mistaken this concept is not wrapping a `Registry` in the other implementations, but rather the other way around.

mxinden (commented):
Couple of comments. Thanks for the work here. More examples are very much appreciated.

examples/actix-web.rs
oriontvv: Good point, agree.

examples/actix-web.rs
oriontvv: Reasonable, please check out a [comment](https://github.com/prometheus/client_rust/issues/53#issuecomment-1079764522)

examples/actix-web.rs
oriontvv: I tried to avoid Mutex here in this abstraction, but didn't found the one.. would like to see any suggestions

mxinden: With https://github.com/prometheus/client_rust/issues/57 fixed, is this still needed @oriontvv?

mxinden: With https://github.com/prometheus/client_rust/issues/57 fixed, is this still needed @oriontvv?

oriontvv: Thanks, it looks like the PR fixes this issue, will try to adapt the example in few days.

mxinden:
:wave: @oriontvv still interested in finishing this pull request?

oriontvv:
Yes, I will simplify the example without unnecessary wrapper in few days 👋

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.16.0] - unreleased
 
+### Added
+
+- Example for actix-web framework. See [PR 54].
+
```
mxinden, line 12:
```suggestion

[PR 54]: https://github.com/prometheus/client_rust/pull/54/

```

examples/actix-web.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct AppState {
+    pub registry: Registry,
+}
+
+pub async fn metrics_handler(state: web::Data<Mutex<AppState>>) -> Result<HttpResponse> {
+    let state = state.lock().unwrap();
+    let mut buf = Vec::new();
+    encode(&mut buf, &state.registry)?;
+    let body = std::str::from_utf8(buf.as_slice()).unwrap().to_string();
+    Ok(HttpResponse::Ok().body(body))
```
mxinden, line 39: Would one not need to set the content type?

https://github.com/prometheus/client_rust/blob/6b7a3458b47a406d15f2ae0a678c611969c4645d/examples/tide.rs#L37

examples/actix-web.rs
```diff
@@ -0,0 +1,73 @@
...
+    metrics.inc_requests(Method::Get);
+    format!("okay")
+}
+
+#[actix_web::main]
+async fn main() -> std::io::Result<()> {
+    let metrics = web::Data::new(Metrics {
+        requests: Family::default(),
+    });
+    // We have to wrap it with Mutex because `Registry` because of actix-web behaviour
```
mxinden, line 52:
```suggestion
    // We have to wrap it with Mutex because of actix-web behaviour.
```

Can you either be more specific, or drop the comment?

mxinden (commented):
Very clean. Two more comments. Otherwise this looks good to me.

examples/actix-web.rs
oriontvv: Actually no, my experience shows that it is not required. Everything works well without this detail.

oriontvv:
Sorry for the garbage commits. I'm trying to figure out how to deal with your tricky checks of signing.

examples/actix-web.rs
mxinden: According to the OpenMetrics specification it "MUST" be set. Am I missing something @oriontvv?

https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#overall-structure

mxinden: //CC @RichiH

mxinden:
> I'm trying to figure out how to deal with your tricky checks of signing.

Sorry for the trouble. It is unfortunate that we require this from a legal side.

src/registry.rs
```diff
@@ -57,7 +57,6 @@ use std::borrow::Cow;
 /// #                "# EOF\n";
 /// # assert_eq!(expected, String::from_utf8(buffer).unwrap());
 /// ```
-#[derive(Debug)]
```
mxinden, line 60: Why this change?

examples/actix-web.rs
oriontvv: Okay, I'll add the header. (btw, we use at work without it and everything is fine)

src/registry.rs
oriontvv: Sorry, it was because of unsuccessful conflicts with rebase, fixed

mxinden (approved):
Looks good to me. Just needs the sign-off. Feel free to just squash into a single signed off commit in case that is more convenient for you @oriontvv.

oriontvv:
reopened in #76

PR #58, opened 2022-03-29
Replace `SendEncodeMetric` with `SendSyncEncodeMetric`

### LINKED ISSUE #57: Sharing dynamically dispatched registry in web framework state
I may be missing something but it seems like it's not possible to use the `Registry` with the default `M = SendEncodeMetric` type as shared state in web frameworks (e.g. `tide::with_state`), because those require that the state is all `Sync` but `SendEncodeMetric` is not `Sync`. If I [update the `tide` example to use a dynamically dispatched `Registry`](https://github.com/sd2k/client_rust/commit/855f4ab5fa68caec27edbab5c8a2d47c39920010) then it fails to compile:

```
✦ ❯ cargo c --example tide
error[E0277]: `(dyn SendEncodeMetric + 'static)` cannot be shared between threads safely
   --> examples/tide.rs:25:36
    |
25  |       let mut app = tide::with_state(State {
    |  ___________________----------------_^
    | |                   |
    | |                   required by a bound introduced by this call
26  | |         registry: Arc::new(registry),
27  | |     });
    | |_____^ `(dyn SendEncodeMetric + 'static)` cannot be shared between threads safely
    |
    = help: the trait `Sync` is not implemented for `(dyn SendEncodeMetric + 'static)`
    = note: required because of the requirements on the impl of `Sync` for `Unique<(dyn SendEncodeMetric + 'static)>`
    = note: required because it appears within the type `Box<(dyn SendEncodeMetric + 'static)>`
    = note: required because it appears within the type `(Descriptor, Box<(dyn SendEncodeMetric + 'static)>)`
    = note: required because of the requirements on the impl of `Sync` for `Unique<(Descriptor, Box<(dyn SendEncodeMetric + 'static)>)>`
    = note: required because it appears within the type `alloc::raw_vec::RawVec<(Descriptor, Box<(dyn SendEncodeMetric + 'static)>)>`
    = note: required because it appears within the type `Vec<(Descriptor, Box<(dyn SendEncodeMetric + 'static)>)>`
    = note: required because it appears within the type `Registry`
    = note: required because of the requirements on the impl of `std::marker::Send` for `Arc<Registry>`
note: required because it appears within the type `State`
   --> examples/tide.rs:59:8
    |
59  | struct State {
    |        ^^^^^
note: required by a bound in `with_state`
   --> /Users/ben/.cargo/registry/src/github.com-1ecc6299db9ec823/tide-0.16.0/src/lib.rs:153:20
    |
153 |     State: Clone + Send + Sync + 'static,
    |                    ^^^^ required by this bound in `with_state`

<several related errors omitted>
```

It's fixed by adding a `Sync` bound to `SendEncodeMetric` but it feels like this is a common use case so I imagine it's come up before and I'm doing something silly - am I?

Thanks!

sd2k:
This will allow the default `Registry` to be used in multi-threaded
environments (e.g. web servers) where the registry needs to be shared
between threads.

Closes #57.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.16.0] - unreleased
+
+### Changed
+
+- Require `Registry` default generic type `SendEncodeMetric` to be `Sync`. See [PR 58].
+
+[PR 58]: https://github.com/prometheus/client_rust/pull/58
+
 ## [0.15.1] - 2022-02-04
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus-client"
-version = "0.15.1"
+version = "0.16.0"
 authors = ["Max Inden <mail@max-inden.de>"]
 edition = "2018"
 description = "Open Metrics client library allowing users to natively instrument applications."
--- a/src/encoding/text.rs
+++ b/src/encoding/text.rs
@@ -404,11 +404,11 @@ impl EncodeMetric for Box<dyn EncodeMetric> {
     }
 }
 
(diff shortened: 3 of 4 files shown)


mxinden (approved):
Thanks @sd2k!

Issue #61, opened 2022-05-11
Add default buckets for Histogram metric


luoxiaoyun-hw:
Histogram metric missed default implementation，wo can add default buckets.

doehyunbaek:
Maybe add default as in client_golang as it is the most popular client implementation?

https://github.com/prometheus/client_golang/blob/66837e3298bdc57a828794c23bacb253ed8c04cd/prometheus/histogram.go#L68

I can contribute a PR as it seems not too hard if you agree.

luoxiaoyun-hw:
we can refer to the client_golang and client_java to provide default buckets.

mxinden:
I would rather call this out in the documentation on `Histogram`, providing an example with the above mentioned default values. I am not in favor of adding a `Default` implementation that only works for some use-cases, here network latency measurements. I find this confusing for newcomers. Experts will be familiar enough to read the example and thus configure it themselves.

> I can contribute a PR as it seems not too hard if you agree.

Improvements to the documentation would be appreciated.

mxinden:
With https://github.com/prometheus/client_rust/pull/63 merged, I am closing here. Feel free to follow up with further comments.

PR #63, opened 2022-05-14
src/metrics/histogram.rs Document no default buckets

doehyunbaek:
As suggested and discussed in https://github.com/prometheus/client_rust/issues/61,
this commit adds a comment that describes unavailability of default
values for buckets.

I thought about providing test case that demonstrates using iterator for buckets as arguments. But thought that would be a bit overkill. If you think it would help let me know.

### DIFF
--- a/src/metrics/histogram.rs
+++ b/src/metrics/histogram.rs
@@ -8,6 +8,8 @@ use std::iter::{self, once};
 use std::sync::{Arc, Mutex, MutexGuard};
 
 /// Open Metrics [`Histogram`] to measure distributions of discrete events.
+/// We do not provide default value for buckets, use predefined buckets
+/// functions or provide your own iterator to argument as constructor.
 ///
 /// ```
 /// # use prometheus_client::metrics::histogram::{Histogram, exponential_buckets};
--- a/src/metrics/histogram.rs
+++ b/src/metrics/histogram.rs
@@ -8,8 +8,17 @@ use std::iter::{self, once};
 use std::sync::{Arc, Mutex, MutexGuard};
 
 /// Open Metrics [`Histogram`] to measure distributions of discrete events.
-/// We do not provide default value for buckets, use predefined buckets
-/// functions or provide your own iterator to argument as constructor.
+///
+/// [`Histogram`] does not implement [`Default`], given that the choice of bucket values depends on the situation [`Histogram`] is used in. As an example, to measure HTTP request latency, the values suggested in the Golang implementation might work for you:
+///
+/// ```
+/// # use prometheus_client::metrics::histogram::Historgam;
+/// // Default values from go client(https://github.com/prometheus/client_golang/blob/5d584e2717ef525673736d72cd1d12e304f243d7/prometheus/histogram.go#L68)
+/// let histogram = Histogram::new(IntoIterator::into_iter([
+///    0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0,
+/// ]));
+/// histogram.observe(4.2);
+/// ```
 ///
 /// ```
 /// # use prometheus_client::metrics::histogram::{Histogram, exponential_buckets};
--- a/src/metrics/histogram.rs
+++ b/src/metrics/histogram.rs
@@ -21,7 +21,7 @@ use std::sync::{Arc, Mutex, MutexGuard};
 /// Golang implementation might work for you:
 ///
 /// ```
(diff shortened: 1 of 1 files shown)


mxinden:
> I thought about providing test case that demonstrates using iterator for buckets as arguments. But thought that would be a bit overkill. If you think it would help let me know.

I think a doc test showcasing the default values from the Go implementation, with a link to the Go code would be great.

src/metrics/histogram.rs
```diff
@@ -8,6 +8,8 @@ use std::iter::{self, once};
 use std::sync::{Arc, Mutex, MutexGuard};
 
 /// Open Metrics [`Histogram`] to measure distributions of discrete events.
+/// We do not provide default value for buckets, use predefined buckets
+/// functions or provide your own iterator to argument as constructor.
```
mxinden, line 12:
```suggestion
///
/// [`Histogram`] does not implement [`Default`], given that the choice of bucket values depends on the situation [`Histogram`] is used in. As an example, to measure HTTP request latency, the values suggested in the Golang implementation might work for you:
```

doehyunbaek:
Was a little sad I could not use array.into_iter() but used IntoIterator::into_iter(array) instead.

Thought about asserting with Histogram struct for doctest but seemed gnarly and settled for simple .observe() call as exponential_buckets example.

src/metrics/histogram.rs
```diff
@@ -9,6 +9,17 @@ use std::sync::{Arc, Mutex, MutexGuard};
 
 /// Open Metrics [`Histogram`] to measure distributions of discrete events.
 ///
+/// [`Histogram`] does not implement [`Default`], given that the choice of bucket values depends on the situation [`Histogram`] is used in. As an example, to measure HTTP request latency, the values suggested in the Golang implementation might work for you:
```
mxinden, line 12: How about moving this section below the vanilla doc example (line 12 / 23)?

mxinden (commented):
Could you wrap the lines?

mxinden:
> Was a little sad I could not use array.into_iter() but used IntoIterator::into_iter(array) instead.

Oh, haven't looked deeply into this. Shouldn't `array.into_iter()` work? Does type inference not work here?

https://doc.rust-lang.org/std/primitive.array.html#impl-IntoIterator

doehyunbaek:
I should have explained more specifically.. Sorry for that.

Arrays have [special behavior](https://doc.rust-lang.org/std/primitive.array.html#:~:text=Arrays%20have%20special%20behavior%20when%20calling%20.into_iter()%20prior%20to%20the%202021%20edition%20%E2%80%93%20see%20the%20array%20Editions%20section%20for%20more%20information.) prior to. 2021 [edition](https://doc.rust-lang.org/std/primitive.array.html#editions).

If we use 2021 edition, array.into_iter would work, if not, we should resort to Intoiterator::into_iter(array) to iterate by value.

doehyunbaek:
Done!

mxinden:
> I should have explained more specifically.. Sorry for that.
> 
> Arrays have [special behavior](https://doc.rust-lang.org/std/primitive.array.html#:~:text=Arrays%20have%20special%20behavior%20when%20calling%20.into_iter()%20prior%20to%20the%202021%20edition%20%E2%80%93%20see%20the%20array%20Editions%20section%20for%20more%20information.) prior to 2021 [edition](https://doc.rust-lang.org/std/primitive.array.html#editions).
> 
> If we use 2021 edition, array.into_iter would work, if not, we should resort to Intoiterator::into_iter(array) to iterate by value.

Oh, right. Thanks for the links. Given that [editions don't split the ecosystem](https://doc.rust-lang.org/edition-guide/editions/index.html#editions-do-not-split-the-ecosystem) what do you think of moving `prometheus-client` to edition `2021` in a separate pull request?

Sorry for derailing your initial proposal here. While not directly related, I think this is very valuable.

src/metrics/histogram.rs
```diff
@@ -14,6 +14,20 @@ use std::sync::{Arc, Mutex, MutexGuard};
...
 /// histogram.observe(4.2);
 /// ```
+///
+/// [`Histogram`] does not implement [`Default`], given that the choice of
+/// bucket values depends on the situation [`Histogram`] is used in. As an
+/// example, to measure HTTP request latency, the values suggested in the
+/// Golang implementation might work for you:
+///
+/// ```
+/// # use prometheus_client::metrics::histogram::Historgam;
```
mxinden, line 24:
```suggestion
/// # use prometheus_client::metrics::histogram::Histogram;
```

Issue #64, opened 2022-05-23
Cargo.toml: Update to Rust edition 2021


mxinden:
As discussed in https://github.com/prometheus/client_rust/pull/63, we should upgrade to Rust edition 2021.

Tracking this here for the sake of completeness.

//CC @doehyunbaek

PR #65, opened 2022-05-26
Update to 2021 edition

doehyunbaek:
I used 2021 edition cargo fix tool (cargo fix --edition) and there was no code change in the repo. Though I am pretty confident in Rust team's ability to make good refactoring tools, I went through changes in the [edition guide](https://doc.rust-lang.org/edition-guide/rust-2021/index.html) and there seemed to be no problem.

One little thing I'm worried about is the message I got when I ran fix tool, which is also pasted in commit description, which is
```console
When building the following dependencies, the given features will no longer be used:
  syn v1.0.95 (as host dependency) removed features: extra-traits, full, visit, visit-mut

The following differences only apply when building with dev-dependencies:
  standback v0.2.17 (as host dependency) removed features: std
  syn v1.0.95 (as host dependency) removed features: visit
```

would you mind taking a look at this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.17.0]
+## [0.17.0] - unreleased
 
 ### Changed
 


doehyunbaek:
closes #64

mxinden (commented):
Thanks!

Just needs:

- Version bump in `Cargo.toml` to `v0.17.0`.
- Changelog entry in `CHANGELOG.md`.
- Edition update to `2021` in `derive-text-encode/Cargo.toml`.
- Version bump in `derive-text-encode/Cargo.toml`.
- Update of `prometheus-client-derive-text-encode` in root level `Cargo.toml` to new version.

doehyunbaek:
Applied suggestion here[1bca5bc](https://github.com/prometheus/client_rust/pull/65/commits/1bca5bc46b9502ce04c2a43bfcd25d3e83cd8b60).

mxinden (approved):
:rocket: perfect! Thanks for the help.

Issue #51, opened 2022-03-09
Text encoding has inconsistent metric names


08d2:

08d2:
It is somewhat surprising that the metric name provided to counter constructors will have `_total` appended to it automatically. It would be cool if that were better documented, and ideally only happened if the provided metric name didn't already end in `_total`.

mxinden:
> The `HELP` and `TYPE` lines should use `my_counter_total`, not `my_counter`.

Not according to the OpenMetrics specification:

https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#counter-1





> It is somewhat surprising that the metric name provided to counter constructors will have `_total` appended to it automatically. It would be cool if that were better documented, and ideally only happened if the provided metric name didn't already end in `_total`.

Good idea. For the former, would you mind extending the documentation @08d2? For the latter, let's move the discussion to https://github.com/prometheus/client_rust/issues/52.

08d2:
> Not according to the OpenMetrics specification

I don't know anything about OpenMetrics, really, and I don't consider it any kind of authority. I can observe that the "_created" thing described in your link is not valid for Prometheus, though.

mxinden:
> AFAIK, this is not compatible with the Prometheus exposition format.

Note that you need to set the right content type in order for Prometheus parse the output of a scrape in the Open Metrics format.

https://github.com/prometheus/client_rust/blob/69e6674064c08630b7b397a58d6e99279d8d98ce/examples/tide.rs#L37

nox:
We have ~100 unsuffixed metrics at work and we are trying to move away from `prometheus` and to switch to `prometheus-client`, this automatic suffixing is an issue, we don't want to change dozens of dashboards because a library forces counter renames on us. Would you accept a PR adding an `LegacyUnsuffixedCounter<A>` type to this crate?

08d2:
> Note that I would see this as a small helper to enable folks to migrate to the OpenMetrics format eventually.

OpenMetrics remains an experiment, not a standard. Prometheus client libraries should default to the core exposition formats, and leave OpenMetrics as an opt-in alternative.

PR #66, opened 2022-06-02
Encode suffix only when it is not presented

quangIO:
This is a temporary workaround for #51, which should work most of the time, with no significantly changes in the API. OpenMetrics format should still works. If users need to use Prometheus counter, then they must explicitly append "_total" to the metric name.


mxinden:
If I am not mistaken, this would allow for the following counter metric output, which is not correct according to the OpenMetrics specification:

```
# TYPE foo_total counter
# HELP foo_total Total number of foo events.
foo_total 17.0
```

Also see https://github.com/prometheus/client_rust/issues/51#issuecomment-1146066161 in order to tell Prometheus to parse the output in OpenMetrics format.

quangIO:
> If I am not mistaken, this would allow for the following counter metric output, which is not correct according to the OpenMetrics specification:
> 
> ```
> # TYPE foo_total counter
> # HELP foo_total Total number of foo events.
> foo_total 17.0
> ```
> 
> Also see [#51 (comment)](https://github.com/prometheus/client_rust/issues/51#issuecomment-1146066161) in order to tell Prometheus to parse the output in OpenMetrics format.

Yes, you are correct. This behaviour is intended is this PR. But like you said, setting `application/openmetrics-text` as content-type is a much better one. So I will close this one.

08d2:
OpenMetrics is not authoritative. Whenever there is a conflict between Prometheus and OpenMetrics, Prometheus should win.

Issue #40, opened 2022-01-24
Implement Summary metric


folex:
Hi! Thank your for the project :)

I found myself missing an implementation of the Summary metric, so decided to file an issue in case anyone (maybe myself) decides to contribute an implementation.

----

Open Metrics spec defines a metric type that computes quantiles locally on the client: [Summary](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#summary).

It's quite useful if you want to learn/discover how a system behaves, especially if you don't have much data a-priori. In that sense, Summary is dual to Histogram - both can be used to understand data distribution (e.g. latency data), but with different use-cases and tradeoffs. 

Good overview on the differences between Summary and Histogram metrics is given in Prometheus doc https://prometheus.io/docs/practices/histograms/

mxinden:
> Thank your for the project :)

Glad that it is useful. :heart: 

> I found myself missing an implementation of the Summary metric, so decided to file an issue in case anyone (maybe myself) decides to contribute an implementation.

Would be good to support the `Summary` metric type. Contributions are most certainly welcome!

Whoever is picking this up, let me know if you need any help.

palash25:
@mxinden hi i would like to work on this could you assign the issue to me?

mxinden:
Done. Thanks @palash25. Let me know in case you need any help.

palash25:
hi @mxinden do you have any preference on what crate to use for the underlying quantile algorithm? I found one that implements CKMS would it be ok to use this?  https://github.com/blt/quantiles

mxinden:
@palash25 unfortunately I don't have any experience with quantile algorithms, neither in general nor in Rust. Thus no preference. Sorry.

> I found one that implements CKMS would it be ok to use this? https://github.com/blt/quantiles

Looks fine to me.

palash25:
hi @mxinden sorry for the multiple pings but can you please take a look at this  https://github.com/prometheus/client_rust/pull/67#issuecomment-1407467095  ? i updated the PR

zhangtianhao:
#249

PR #67, opened 2022-06-26
Add Summary metric type

palash25:
Not ready for review yet, just opening a draft for visibility

Closes: #40

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ dtoa = "1.0"
 itoa = "1.0"
 owning_ref = "0.4"
 prometheus-client-derive-text-encode = { version = "0.3.0", path = "derive-text-encode" }
+quantiles = "0.7.1"
 
 [dev-dependencies]
 async-std = { version = "1", features = ["attributes"] }
--- a/src/metrics/summary.rs
+++ b/src/metrics/summary.rs
@@ -1,0 +1,117 @@
+//! Module implementing an Open Metrics histogram.
+//!
+//! See [`Summary`] for details.
+
+use super::{MetricType, TypedMetric};
+//use owning_ref::OwningRef;
+//use std::iter::{self, once};
+use std::sync::{Arc, Mutex};
+
+use quantiles::ckms::CKMS;
+
+/// Open Metrics [`Summary`] to measure distributions of discrete events.
+pub struct Summary {
+    target_quantile: Vec<f64>,
+    target_error: f64,
+    max_age_buckets: u64,
+    max_age_seconds: u64,
+    inner: Arc<Mutex<InnerSummary>>,
+}
+
+impl Clone for Summary {
+    fn clone(&self) -> Self {
+        Summary {
+            target_quantile: self.target_quantile.clone(),
+            target_error: self.target_error,
+            max_age_buckets: self.max_age_buckets,
+            max_age_seconds: self.max_age_seconds,
+            inner: self.inner.clone(),
+        }
+    }
+}
+
+pub(crate) struct InnerSummary {
+    sum: f64,
+    count: u64,
+    quantile_streams: Vec<CKMS<f64>>,
+    // head_stream is like a cursor which carries the index
+    // of the stream in the quantile_streams that we want to query
+    head_stream: u64,
+}
+
+impl Summary {
+    pub fn new(max_age_buckets: u64, max_age_seconds: u64, target_quantile: Vec<f64>, target_error: f64) -> Self {
+        let mut streams: Vec<CKMS<f64>> = Vec::new();
+        for _ in 0..max_age_buckets {
+            streams.push(CKMS::new(target_error));
+        }
+
+        Summary{
+            max_age_buckets,
+            max_age_seconds,
+            target_quantile,
+            target_error,
+            inner: Arc::new(Mutex::new(InnerSummary {
+                sum: Default::default(),
+                count: Default::default(),
+                quantile_streams: streams,
+                head_stream: 0,
+            }))
+        }
+    }
+
+    pub fn observe(&mut self, v: f64) {
+        let mut inner = self.inner.lock().unwrap();
+        inner.sum += v;
+        inner.count += 1;
+
+        // insert quantiles into all streams/buckets.
+        for stream in inner.quantile_streams.iter_mut() {
+            stream.insert(v);
+        }
+    }
+
+    pub fn get(&self) -> (f64, u64, Vec<(f64, f64)>) {
+        let inner = self.inner.lock().unwrap();
+        let sum = inner.sum;
+        let count = inner.count;
+        let head = inner.head_stream;
+        let mut quantile_values: Vec<(f64, f64)> = Vec::new();
+
+        // TODO: add stream rotation
+        for q in self.target_quantile.iter() {
+            match inner.quantile_streams[head as usize].query(*q) {
+                Some((_, v)) => quantile_values.push((*q, v)),
+                None => continue, // TODO fix this
+            };
+        }
+        (sum, count, quantile_values)
+    }
+}
+
+// TODO: should this type impl Default like Counter?
+
+impl TypedMetric for Summary {
+    const TYPE: MetricType = MetricType::Summary;
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn basic() {
+        let mut summary = Summary::new(5, 10, vec![0.5, 0.9, 0.99], 0.01);
+        summary.observe(5.0);
+        summary.observe(15.0);
+        summary.observe(25.0);
+
+        let (s, c, q) = summary.get();
+        assert_eq!(45.0, s);
+        assert_eq!(3, c);
+
+        for elem in q.iter() {
+            println!("Vec<{}, {}>", elem.0, elem.1);
+        }
+    }
+}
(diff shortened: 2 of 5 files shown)


palash25:
sorry i completely forgot about this PR @mxinden can you please take a look

palash25:
i will add some comments in the `summary.rs` file and some doctests shortly

gmossessian:
@palash25 do you need a hand here? I'm also wanting this metric, I started implementing it myself before I saw your PR. Looks like you still need to implement changes to `proto.rs`, can I help with this?

@mxinden in your opinion is anything besides this needed here before releasing?

palash25:
oh you are right, i missed the proto.rs file thanks for pointing it out, i will try to make the remaining changes in the next 2 days, if I can't i will ping you for help.

palash25:
moving this back to draft till i make the missing changes

mxinden:
@palash25 #105 introduced a fairly large change. I am sorry for the conflict. Let me know in case you need any help updating.

palash25:
sorry for the late response @mxinden the notification got buried. I tried to make the changes according to the new encoding API but I am stuck here.

Right now my code errors with this

```
error[E0282]: type annotations needed
   --> src/encoding.rs:145:15
    |
145 |             e.encode_summary(sum, count, quantiles)
    |               ^^^^^^^^^^^^^^ cannot infer type of the type parameter `S` declared on the associated function `encode_summary`
    |
help: consider specifying the generic argument
    |
145 |             e.encode_summary::<S>(sum, count, quantiles)
    |                             +++++

```


if I do as the compiler says and add this annotation ` e.encode_summary::<S>(sum, count, quantiles)` all the tests start passing but then my code looks a bit inconsistent in comparison to other `encode_*` methods in the file. 

Do you know what I might be doing wrong here?

palash25:
Hi @mxinden  could you take a look at the comment? ☝🏽

src/metrics/summary.rs
```diff
@@ -0,0 +1,165 @@
+//! Module implementing an Open Metrics histogram.
```
mxinden, line 1:
```suggestion
//! Module implementing an Open Metrics summary.
```

src/metrics/summary.rs
```diff
@@ -0,0 +1,165 @@
...
+//!
+//! See [`Summary`] for details.
+
+use crate::encoding::{EncodeMetric, MetricEncoder};
+
+use super::{MetricType, TypedMetric};
+//use owning_ref::OwningRef;
+//use std::iter::{self, once};
+use std::sync::{Arc, Mutex};
```
mxinden, line 10: We import `parking_lot` anyways, thus I suggest using `Mutex` from there.

mxinden (commented):
Sorry for the delay here. Implementation of encoding in text and protobuf looks good to me. Couple minor comments.

Before I dive deeper into the implementation of Summary, can you expand on why you chose CKMS and how it differs to the [golang implementation](https://github.com/prometheus/client_golang/blob/main/prometheus/summary.go)?

mxinden:
> Before I dive deeper into the implementation of Summary, can you expand on why you chose CKMS and how it differs to the [golang implementation](https://github.com/prometheus/client_golang/blob/main/prometheus/summary.go)?

@beorn7 given that you wrote most of the client-golang Summary implementation and given your involvement in https://github.com/OpenObservability/OpenMetrics/pull/256 I would appreciate your input here. Do you have opinions on which quantile algorithm to use?

palash25:
so about using CKMS, i went through both go and java libraries, the underlying quantile library that go uses has some inacurracies mentioned here https://github.com/beorn7/perks/blob/master/quantile/stream.go#L174  so I looked at the java client implementation and it seemed to be using a CKMS library https://github.com/prometheus/client_java/blob/c28b901225e35e7c1df0eacae8b58fdfbb390162/simpleclient/src/main/java/io/prometheus/client/TimeWindowQuantiles.java#L14  at the same time I was able to find a rust crate implementing CKMS so I just went ahead with it.

Thats all I can remember since it was a long time ago

> how it differs to the [golang implementation](https://github.com/prometheus/client_golang/blob/main/prometheus/summary.go?rgh-link-date=2023-02-24T18%3A01%3A19Z)?


Not sure I remember the differences but this is what I wrote in my notes at the time which seems to be a similarity between the go library and the rust crate: **"Rust crate uses the LowBiased invariant defined in the golang one   [quantiles/store.rs at master · blt/quantiles · GitHub](https://github.com/blt/quantiles/blob/master/src/ckms/store.rs#L25)
"**

beorn7:
tl;dr: Use whatever algorithm you find appropriate. It would require a lot of research to assess all the algorithms out there and pick the "best" one for client_rust.

The algorithm client_golang uses was picked even before my time. I don't have strong opinions about it. It's perfectly possible that there are "better" algorithms out there. It has been a hot topic of research for a while, so there are many algorithms available. It's of course also hard to decide about the "best" algorithm because that surely depends on your use case.

There was an attempt to add a now algorithm or replace the existing one, see https://github.com/prometheus/client_golang/pull/859 . It was closed for lack of follow-up when we tried to discuss the aspects in which it would be a better algorithm.

I do believe that the algorithm in client_golang is implemented correctly (i.e. as it is described in the paper). client_golang uses my fork, which includes fixes. The `Merge` method @palash25 has mentioned above is still broken, and it cannot be fixed, because the used algorithm doesn't support merging. But that method is also not used in client_golang.

src/metrics/summary.rs
```diff
@@ -0,0 +1,164 @@
...
+                last_rotated_timestamp,
+            }))
+        }
+    }
+
+    /// Observe the given value.
+    pub fn observe(&self, v: f64) {
+        self.rotate_buckets();
+
+        let mut inner = self.inner.write();
```
mxinden, line 85: Both `rotate_buckets` and later on `observe` itself grab the write lock. The lock will be unlocked in between. Thus another thread could potentially execute in between. Does that violate any consistency guarantees. In other words does `observe` depend on `rotate_buckets` to be called right before?

See similar race condition in https://github.com/prometheus/client_rust/pull/102.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ parking_lot = "0.12"
 prometheus-client-derive-encode = { version = "0.4.1", path = "derive-encode" }
 prost = { version = "0.11.0", optional = true }
 prost-types = { version = "0.11.0", optional = true }
+quantiles = "0.7.1"
```
mxinden, line 27: I would guess that most users of this library don't need the `Summary` type. What do you think of feature flagging the type as well as the `quantiles` dependency. The latter will help keeping ones dependency tree lean, thus compile times short.

src/metrics/summary.rs
```diff
@@ -0,0 +1,164 @@
+//! Module implementing an Open Metrics summary.
+//!
+//! See [`Summary`] for details.
+
+use crate::encoding::{EncodeMetric, MetricEncoder};
+
+use super::{MetricType, TypedMetric};
+use parking_lot::RwLock;
+use std::sync::Arc;
+use std::time::{Duration, Instant};
```
mxinden, line 10: `std::time::Instant` is problematic with wasm. Consider using https://crates.io/crates/instant instead.

src/metrics/summary.rs
```diff
@@ -0,0 +1,164 @@
...
+        inner.count += 1;
+
+        // insert quantiles into all streams/buckets.
+        for stream in inner.quantile_streams.iter_mut() {
+            stream.insert(v);
+        }
+    }
+
+    /// Retrieve the values of the summary metric.
+    pub fn get(&self) -> (f64, u64, Vec<(f64, f64)>) {
```
mxinden, line 96:
```suggestion
    pub(crate) fn get(&self) -> (f64, u64, Vec<(f64, f64)>) {
```

Is this needed outside of the encode module?

src/metrics/summary.rs
```diff
@@ -0,0 +1,164 @@
...
+        self.rotate_buckets();
+
+        let inner = self.inner.read();
+        let sum = inner.sum;
+        let count = inner.count;
+        let mut quantile_values: Vec<(f64, f64)> = Vec::new();
```
mxinden, line 102: Instead of allocating, you could consider returning an `Iterator` instead. Though likely premature optimization without any proof that this is an issue.

src/metrics/summary.rs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+#[derive(Debug)]
+pub(crate) struct InnerSummary {
+    sum: f64,
+    count: u64,
+    quantile_streams: Vec<CKMS<f64>>,
+    // head_stream is like a cursor which carries the index
+    // of the stream in the quantile_streams that we want to query.
+    head_stream_idx: u64,
```
mxinden, line 45:
```suggestion
    head_stream_idx: usize,
```

Why not use a `usize` here as it is an index into a datastructure? You compare it to `max_age_buckets`, but that could be a `usize` as well.

mxinden (commented):
Thanks @palash25 and @beorn7 for the input :pray:

Overall this looks good to me. I have a couple of smaller comments, nothing major. Let me know in case you need help with the feature flagging.

Cargo.toml
cz-cube: Are we blocked by this issue?

src/metrics/summary.rs
mxinden: @cz-cube definitely needs some investigation.

palash25:
I am sorry for the inactivity here. I have not been in the best of health for the past few months. Please feel free to take over this PR and make whatever changes are required. You can even close and let someone else have a crack at it if they want to start from scratch.

PR #70, opened 2022-07-08
Add with_prefix to Registry

Victor-N-Suadicani:
Adds the ability to create a registry with a prefix.

This is actually already possible through the below hack but this makes it a lot easier and neater.

```rust
let prefixed_registry = {
    let prefixed_registry = registry.sub_registry_with_prefix(prefix);
    std::mem::take(prefixed_registry)
};
```

This is useful to set an application-wide prefix on the single registry for all your metrics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updates to Rust 2021 Edition. See [PR 65].
 
 ### Added
-- Added a `with_prefix` method to `Registry` to allow initializing a registry with a prefix. See [PR 70]
+- Added a `with_prefix` method to `Registry` to allow initializing a registry with a prefix. See [PR 70].
 
 ### Removed
 - Remove `Add` trait implementation for a private type which lead to compile time conflicts with existing `Add` implementations e.g. on `String`. See [PR 69].
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,15 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.17.0] - unreleased
 
 ### Changed
-
 - Updates to Rust 2021 Edition. See [PR 65].
 
-### Removed
+### Added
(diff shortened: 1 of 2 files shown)


src/registry.rs
```diff
@@ -77,6 +77,14 @@ impl<M> Default for Registry<M> {
 }
 
 impl<M> Registry<M> {
+    /// Creates a new default registry with the given prefix.
+    pub fn with_prefix(prefix: impl Into<String>) -> Self {
```
mxinden, line 81:
```suggestion
    pub fn new_with_prefix(prefix: impl Into<String>) -> Self {
```

This would be more intuitive for me. What do you think @Victor-N-Suadicani ?

src/registry.rs
```diff
@@ -77,6 +77,14 @@ impl<M> Default for Registry<M> {
 }
 
 impl<M> Registry<M> {
+    /// Creates a new default registry with the given prefix.
```
mxinden, line 80:
```suggestion
    /// Creates a new default [`Registry`] with the given prefix.
```

mxinden (commented):
Just two comments. Thanks for sending in a pull request!

Would you mind adding a changelog entry?

src/registry.rs
Victor-N-Suadicani: The convention in Rust is to just use `with_something` - check for instance `with_capacity` on `Vec` or `with_hasher` on `HashMap` or just search the `std` docs for `with_` for lots of examples.

CHANGELOG.md
```diff
@@ -7,15 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.17.0] - unreleased
 
 ### Changed
-
 - Updates to Rust 2021 Edition. See [PR 65].
 
-### Removed
+### Added
+- Added a `with_prefix` method to `Registry` to allow initializing a registry with a prefix. See [PR 70]
```
mxinden, line 13:
```suggestion
- Added a `with_prefix` method to `Registry` to allow initializing a registry with a prefix. See [PR 70].
```

mxinden (approved):
Great. Thanks!

PR #71, opened 2022-07-08
Add Debug impls and fix clippy warnings

Victor-N-Suadicani:
Lots of public types are missing `Debug` impls. I've added most of the easy Debug impls and also added `#![warn(missing_debug_implementations)]` to the top level.

There are also quite a few Clippy warnings that have been fixed.


src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![forbid(unsafe_code)]
 #![deny(unused)]
 #![deny(dead_code)]
+#![warn(missing_debug_implementations)]
```
mxinden, line 4: I am a bit undecided. I don't think we should be adding `Debug` implementations to types by default as it slows compile time and bloats the compiled binary.

Do you feel strongly about this? What do you think of removing this line but keeping the new `derive` statements below?

```suggestion
```

src/encoding/text.rs
```diff
@@ -616,7 +620,7 @@ mod tests {
     fn encode_counter() {
         let counter: Counter = Counter::default();
         let mut registry = Registry::default();
-        registry.register("my_counter", "My counter", counter.clone());
+        registry.register("my_counter", "My counter", counter);
```
mxinden, line 623: What do you think of adding a commit here that alters `.github/workflow` to lint the test code as well?

mxinden (commented):
Thanks for the work here @Victor-N-Suadicani!

src/lib.rs
Victor-N-Suadicani: This only lints for public types I believe. So it's not every type.

But I don't feel _super_ strongly about it. Though I think it is a good idea.

Your call :)

src/encoding/text.rs
Victor-N-Suadicani: I added `--workspace --all-targets` to the arguments of the clippy action. That should include all packages and include tests.

src/lib.rs
mxinden: > This only lints for public types I believe. So it's not every type.

Good point. Thanks. Let's keep it as is then.

mxinden (commented):
Looks good to me. Only missing a changelog entry for the added `Debug` implementations.

mxinden (approved):
Thanks @Victor-N-Suadicani. I will cut a new release soon.

Issue #73, opened 2022-07-19
Combining registries?


vladvasiliu:
Is there an elegant way of combining multiple registries in a single encode output?

An example of a use case would be writing an exporter for metrics from an external system, as opposed to the own application's metrics.

The idea is that new metric families would be created and populated on each scrape. But there could also be exporter-related metrics that would be constant between scrapes, such as an info, etc.

One way of doing this right now would be to register the exporter's metrics to a new, per-scrape Registry, together with the scrape metrics and encode that. This approach seems somewhat verbose.

A more ergonomic approach would be for the encode function to take some iterable of Families and encode that, which is basically what the Registry is from its point of view.

mxinden:
Would the `Collector` pattern be helpful here?

Proposed in https://github.com/prometheus/client_rust/issues/49

Further designed in https://github.com/prometheus/client_rust/issues/29#issuecomment-1172877769

vladvasiliu:
Yeah, I think that could be helpful.

I'd initially glossed over these, seeing how at first they were about "custom encoders", but upon closer reading that's pretty much what I'm after.

mxinden:
Great. Note @vladvasiliu that help here is very much appreciated.

vladvasiliu:
Sure, I'll see if I can come up with something over the weekend.

Issue #29, opened 2021-12-14
Implementing a process collector


gagbo:
Hello,

I recently stumbled upon https://github.com/tikv/rust-prometheus/issues/392 and now I'm preparing a migration from `rust-prometheus`. I'm wondering how it would be possible to add a collector for process metrics just like what `rust-prometheus` does currently. I'm really inexperienced in those packages, and I don't really see how the `ProcessCollector` thing would translate here, to add some metrics related to the running process.

Can you tell me what would be necessary to add support for this ? I was thinking probably as another crate that exposes a single global function like `crate::export_process_metrics()`; so that crate would have to add timers to run the collection and hope that the timer runs often enough to give a precise measurement when prometheus scrapes the endpoint ?

Regards,
Gerry

mxinden:
Hi Gerry,

As you noted above `open-metrics-client` does not itself support exposition of process metrics today.

> I was thinking probably as another crate

That sounds good. I am open to whether that crate would live in this repository or not. The former would likely make development easier as one can make atomic changes across both crates at once (with one pull request).

Off the top of my head the interface of such crate could look like:

``` rust
fn register_process_metrics(registry: &mut Registry)
```

The crate would register multiple custom metrics, e.g. a gauge metric exposing the number of threads. Those custom metrics would each implement [`EncodeMetric`](https://docs.rs/open-metrics-client/0.13.0/open_metrics_client/encoding/text/trait.EncodeMetric.html) and collect the concrete metric values from the system on `EncodeMetric::encode`.

@gagbo does the above make sense? Would you be interested in contributing such a crate?

gagbo:
If I can find some time it’d be nice yeah, hopefully I’ll be able to find time to dig into this. 

I think I’d like a way to specify running the `encode` function in a separate thread (which would require the registry to be Sync ?), so that all the metrics related calls can be handled by a different core if need be ?

mxinden:
> If I can find some time it’d be nice yeah, hopefully I’ll be able to find time to dig into this.

Great. Let me know in case you need any help!

> I think I’d like a way to specify running the encode function in a separate thread

Can you expand on what you want to optimize?  All metric implementations are synchronized (e.g. `Counter` is just an atomic integer), thus allowing metric recording and metric collection to happen in different threads.

> (which would require the registry to be Sync ?)

[...]

PR #74, opened 2022-07-26
Collector pattern

vladvasiliu:
This is an implementation of a `Collector` trait for collecting all the Metric Families from one or more Registries to encode them all at once.

For example, this allows exporting metrics by combining a "long-running" `Registry` with a per-scrape `Registry`.

This is in response to #29, #49 and #73.

Implementation is rather crude, but the out-facing interface is basically unchaged.

### DIFF
--- a/examples/tide.rs
+++ b/examples/tide.rs
@@ -31,7 +31,7 @@ async fn main() -> std::result::Result<(), std::io::Error> {
     app.at("/metrics")
         .get(|req: tide::Request<State>| async move {
             let mut encoded = Vec::new();
-            encode(&mut encoded, &req.state().registry).unwrap();
+            encode(&mut encoded, req.state().registry.as_ref()).unwrap();
             let response = tide::Response::builder(200)
                 .body(encoded)
                 .content_type("application/openmetrics-text; version=1.0.0; charset=utf-8")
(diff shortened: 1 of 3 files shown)


mxinden:
I am sorry for the late reply. This is a neat solution. I did not think about abstracting a `Registry`.

Unfortunately a user can not use the `Collector` abstraction in combination with the `Registry::sub_registry_with_xxx` mechanism. I would expect this to be useful for users. Would you agree? This is similar to the suggestion in https://github.com/prometheus/client_rust/issues/29#issuecomment-1172877769.

I have to give this more thought. Have you had the chance to play with this approach instead?

vladvasiliu:
> Unfortunately a user can not use the Collector abstraction in combination with the Registry::sub_registry_with_xxx mechanism. I would expect this to be useful for users. Would you agree?

I'd have to get back to looking at this. Right now, it's not clear to me why that wouldn't work. The collector delegates the actual encoding to the `Registry`, so it's the `Registry`'s business to deal with its subregistries, and I haven't touched that. There's a test checking for this, which passes.

https://github.com/prometheus/client_rust/blob/8b7ab5bff2b24ea75350ae5cd584ea341ba621f0/src/encoding/text.rs#L714

I'd have to check to make sure it doesn't somehow sidestep the issue.

> This is similar to the suggestion in https://github.com/prometheus/client_rust/issues/29#issuecomment-1172877769.

That was actually my inspiration for this. The overall idea was to allow `encode` to take a bunch of `Collector`s instead of a `Registry`. But the user shouldn't have to deal with a `Collector` directly, just use the `Registry` as usual.

mxinden:
Cross referencing alternative proposal: https://github.com/prometheus/client_rust/pull/82

mxinden:
Implemented through https://github.com/prometheus/client_rust/pull/82. Let me know what you think. Closing here.

PR #76, opened 2022-07-28
Add actix-web example

### LINKED ISSUE #53: Add example for actix-web framework

oriontvv:
Resolves #53


oriontvv:
I decided to refresh the PR and have made clear squashed signoff changes which are the same as in previous version.

mxinden (approved):
:pray: 

I am sure many newcomers will find this helpful.

Issue #77, opened 2022-08-02
Replace owning_ref


DieracDelta:
Owning ref is (1) seemingly unmaintained since last commit was 2 years ago and (2) [unsound](https://github.com/noamtashma/owning-ref-unsoundness). Something else should probably be used.

AbhijithGanesh:
True, this issue has popped up here in [LibP2P](https://github.com/libp2p/rust-libp2p/issues/2794) and other crates which uses this prometheus client

AbhijithGanesh:
Can I work on patching the issue ? @mxinden

AbhijithGanesh:
The `Owning_Ref` library is used at the following files:
 - [exemplar.rs](https://github.com/prometheus/client_rust/blob/master/src/metrics/exemplar.rs)
 - [family.rs](https://github.com/prometheus/client_rust/blob/master/src/metrics/family.rs)
 - [histogram.rs](https://github.com/prometheus/client_rust/blob/master/src/metrics/histogram.rs)

Histogram relevant files use this module, they cannot be used due to Security reasons! This needs to be patched where the usage of Owning_Ref is dropped!

DieracDelta:
In other issues I see [this](https://docs.rs/reffers/latest/reffers/) being recommended. Haven't had a chance to look too deeply, but perhaps that is a suitable replacement?

mxinden:
> Can I work on patching the issue ? @mxinden

Yes, for sure. Help is very much appreciated.

> In other issues I see [this](https://docs.rs/reffers/latest/reffers/) being recommended. Haven't had a chance to look too deeply, but perhaps that is a suitable replacement?

Can you tell whether it suffers the same unsoundness issues that `owning_ref` does?

---

We need `owning_ref` in _wrapper style_ metrics (e.g. `Family`), which internally use a `Mutex` though want to give users access to a reference of the underlying _wrapped_ metric (e.g. `Counter`). Given that this reference depends on the `MutexGuard`, we need to ensure that the `MutexGuard` lives long enough, i.e. throughout the lifetime of the reference of the _wrapped_ metric. Without keeping the `MutexGuard` alive, we invalidate the properties of the `Mutex` itself, namely mutual exclusive access to the _wrapped_ metric.

Thus far `owning_ref` offered a way to ship both the `MutexGuard` and the reference to the user, thus enforcing the lifetime bound.

Alternative approaches I can see:

[...]

PR #78, opened 2022-08-06
*: Move to `parking_lot` and thus make `owning_ref` obsolete

mxinden:
Before `proemtheus-client` would use the `owning_ref` crate to map the target
of a `std::sync::RwLockReadGuard`. `owning_ref` has multiple unsoundness
issues, see https://rustsec.org/advisories/RUSTSEC-2022-0040.html. Instead of
replacing `owning_ref` with a similar crate, we switch to locking via
`parking_lot` which supports the above mapping natively.

Fixes https://github.com/prometheus/client_rust/issues/77

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.18.0] - unreleased
+
+### Changed
+
+- Use `parking_lot` instead of `std::sync::*`.
+
+  Before `proemtheus-client` would use the `owning_ref` crate to map the target
+  of a `std::sync::RwLockReadGuard`. `owning_ref` has multiple unsoundness
+  issues, see https://rustsec.org/advisories/RUSTSEC-2022-0040.html. Instead of
+  replacing `owning_ref` with a similar crate, we switch to locking via
+  `parking_lot` which supports the above mapping natively.
+
 ## [0.17.0]
 
 ### Changed
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus-client"
-version = "0.17.0"
+version = "0.18.0"
 authors = ["Max Inden <mail@max-inden.de>"]
 edition = "2021"
 description = "Open Metrics client library allowing users to natively instrument applications."
@@ -16,7 +16,7 @@ members = ["derive-text-encode"]
 [dependencies]
 dtoa = "1.0"
 itoa = "1.0"
-owning_ref = "0.4"
+parking_lot = "0.12"
 prometheus-client-derive-text-encode = { version = "0.3.0", path = "derive-text-encode" }
 
 [dev-dependencies]
--- a/src/encoding/text.rs
+++ b/src/encoding/text.rs
@@ -449,7 +449,7 @@ where
 {
     fn encode(&self, encoder: Encoder) -> Result<(), std::io::Error> {
         let (value, exemplar) = self.get();
-        encode_counter_with_maybe_exemplar(value, exemplar.as_ref().as_ref(), encoder)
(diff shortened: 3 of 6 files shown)


mxinden:
I wonder whether moving to `parking_lot` has any implications on this crate's support for WASM.

Gathering the WASM experts here. Do you have any opinions on this? Do we need to pipe through any feature flags? 

@PhilippGackstatter, @thomaseizinger, @mriise, @stebalien

mxinden:
For the record, this patch does not seem to have a significant impact on benchmarks:

```
➜  critcmp master parking_lot
group                                                  master                                 parking_lot
-----                                                  ------                                 -----------
counter family with Vec<(String, String)> label set    1.00    150.9±1.46ns        ? ?/sec    1.02    153.9±3.00ns        ? ?/sec
counter family with custom type label set              1.00     40.4±0.30ns        ? ?/sec    1.11     44.9±4.67ns        ? ?/sec
encode                                                 1.07     26.5±0.16ms        ? ?/sec    1.00     24.8±1.10ms        ? ?/sec

➜  grep -m 1 'model name' /proc/cpuinfo
model name      : Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
```

That said, I don't think the benchmarks are at a point, where I would say that they represent real world use-cases.

thomaseizinger:
> I wonder whether moving to `parking_lot` has any implications on this crate's support for WASM.
> 
> Gathering the WASM experts here. Do you have any opinions on this? Do we need to pipe through any feature flags?
> 
> @PhilippGackstatter, @thomaseizinger, @mriise, @Stebalien

Not sure if I'd consider myself a WASM expert :see_no_evil: 

Do we currently support compiling to WASM? Looking through the [CI config](https://github.com/prometheus/client_rust/blob/c9b42b1801dfd1ef99d8fe3295b46a5458160b8f/.github/workflows/rust.yml#L144-L149), it doesn't seem to be or at least it is not verified? Adding `wasm32-unknown-unknown` should at least give us some confidence here although we should bear in mind that often, things compile to wasm but don't actually work at runtime. To be really sure, we would have to add `wasm_bindgen_test` based tests.

thomaseizinger:
If WASM support is something we want to guarantee, I think we should make a separate PR that adds the target in the CI cross-compile config, merge that first and see if this PR is affected by it.

AbhijithGanesh:
@mxinden and @thomaseizinger , I am a new contributor to this project, I am interested in knowing why we should or shouldn't support WASM ? Why WASM for a client that exports metrics? I am not able to understand the usage of WASM here, can you explain!

thomaseizinger:
> @mxinden and @thomaseizinger , I am a new contributor to this project, I am interested in knowing why we should or shouldn't support WASM ? Why WASM for a client that exports metrics? I am not able to understand the usage of WASM here, can you explain!

WASM is becoming a more and more relevant deployment target and no longer necessarily means "browser" or "web".

See for example:

- https://github.com/fermyon/spin
- https://github.com/lunatic-solutions/lunatic

It is completely feasible that users would want to collect metrics in applications written for either of these two platforms in which case they have to compile the Rust client for prometheus to WASM.

AbhijithGanesh (approved):
LGTM

thomaseizinger:
> If WASM support is something we want to guarantee, I think we should make a separate PR that adds the target in the CI cross-compile config, merge that first and see if this PR is affected by it.

I've made a PR for that here: https://github.com/prometheus/client_rust/pull/79

mxinden:
> Note: It looks like parking_lot actually has WASM support. When compiled without (experimental) WASM-atomic support, it just panics when trying to park, but that's fine.

Adding to this:

> The wasm32-unknown-unknown target is only fully supported on nightly with -C target-feature=+atomics in RUSTFLAGS and -Z build-std passed to cargo. parking_lot will work mostly fine on stable, the only difference is it will panic instead of block forever if you hit a deadlock. Just make sure not to enable -C target-feature=+atomics on stable as that will allow wasm to run with multiple threads which will completely break parking_lot's concurrency guarantees.

https://github.com/Amanieu/parking_lot#nightly-vs-stable

mxinden:
Unless there are any objections, I will merge and release this tomorrow or next week. Thanks for the input everyone!

PR #79, opened 2022-08-08
Add `wasm32` as a cross-compile target

thomaseizinger:


.github/workflows/rust.yml
```diff
@@ -147,6 +147,7 @@ jobs:
           - mipsel-unknown-linux-gnu
           - powerpc-unknown-linux-gnu
           - powerpc64-unknown-linux-gnu
+          - wasm32-unknown-unknown
```
mxinden, line 150:
```suggestion
          - wasm32-unknown-unknown
          - wasm32-unknown-emscripten,
          - wasm32-wasi
```

Would it make sense to test against these targets as well?

mxinden (commented):
:pray: this help is much appreciated!

.github/workflows/rust.yml
thomaseizinger: I think `wasm32-unknown-unknown` is the most restrictive one. If it works for those, it should work for the others too :)

mxinden: Understood. Let's proceed as is then.

mxinden (approved):

PR #82, opened 2022-08-29
src/collector: Introduce Collector abstraction

mxinden:
The `Collector` abstraction allows users to provide additional metrics
and their description on each scrape.

See also:

- https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#hdr-Custom_Collectors_and_constant_Metrics
- https://github.com/prometheus/client_rust/issues/49
- https://github.com/prometheus/client_rust/issues/29

I am opening this up as a "Draft" for early feedback.

### DIFF
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -1,0 +1,28 @@
+//! Metric collector implementation.
+//!
+//! See [`Collector`] for details.
+
+use std::borrow::Cow;
+
+use crate::{
+    registry::{Descriptor, Metric},
+    MaybeOwned,
+};
+
+/// The [`Collector`] abstraction allows users to provide additional metrics and
+/// their description on each scrape.
+///
+/// An example use-case is an exporter that retrieves a set of operating system metrics
+/// ad-hoc on each scrape.
+///
+/// Register a [`Collector`] with a [`Registry`](crate::registry::Registry) via
+/// [`Registry::register_collector`](crate::registry::Registry::register_collector).
+pub trait Collector: std::fmt::Debug + Send + Sync + 'static {
+    /// Once the [`Collector`] is registered, this method is called on each scrape.
+    ///
+    /// Note that the return type allows you to either return owned (convenient)
+    /// or borrowed (performant) descriptions and metrics.
+    fn collect<'a>(
+        &'a self,
+    ) -> Box<dyn Iterator<Item = (Cow<'a, Descriptor>, MaybeOwned<'a, Box<dyn Metric>>)> + 'a>;
+}
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -4,7 +4,10 @@
 
 use std::borrow::Cow;
 
-use crate::{registry::Descriptor, MaybeOwned};
+use crate::{
+    registry::{Descriptor, LocalMetric},
+    MaybeOwned,
+};
 
 /// The [`Collector`] abstraction allows users to provide additional metrics and
 /// their description on each scrape.
@@ -21,11 +24,5 @@ pub trait Collector: std::fmt::Debug + Send + Sync + 'static {
     /// or borrowed (performant) descriptions and metrics.
     fn collect<'a>(
         &'a self,
-    ) -> Box<dyn Iterator<Item = (Cow<'a, Descriptor>, MaybeOwned<'a, Box<dyn Metric>>)> + 'a>;
+    ) -> Box<dyn Iterator<Item = (Cow<'a, Descriptor>, MaybeOwned<'a, Box<dyn LocalMetric>>)> + 'a>;
 }
-
-// TODO: Footgun as it is hard to differentiate with registry::Metric.
-/// Super trait representing an abstract Prometheus metric.
-pub trait Metric: crate::encoding::EncodeMetric + std::fmt::Debug {}
-
-impl<T> Metric for T where T: crate::encoding::EncodeMetric + std::fmt::Debug {}
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -22,6 +22,7 @@ pub trait Collector: std::fmt::Debug + Send + Sync + 'static {
     ///
     /// Note that the return type allows you to either return owned (convenient)
     /// or borrowed (performant) descriptions and metrics.
+    #[allow(clippy::type_complexity)]
     fn collect<'a>(
         &'a self,
     ) -> Box<dyn Iterator<Item = (Cow<'a, Descriptor>, MaybeOwned<'a, Box<dyn LocalMetric>>)> + 'a>;
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -4,10 +4,7 @@
 
 use std::borrow::Cow;
 
-use crate::{
-    registry::{Descriptor, Metric},
-    MaybeOwned,
-};
+use crate::{registry::Descriptor, MaybeOwned};
 
 /// The [`Collector`] abstraction allows users to provide additional metrics and
 /// their description on each scrape.
@@ -26,3 +23,9 @@ pub trait Collector: std::fmt::Debug + Send + Sync + 'static {
         &'a self,
     ) -> Box<dyn Iterator<Item = (Cow<'a, Descriptor>, MaybeOwned<'a, Box<dyn Metric>>)> + 'a>;
 }
+
+// TODO: Footgun as it is hard to differentiate with registry::Metric.
+/// Super trait representing an abstract Prometheus metric.
+pub trait Metric: crate::encoding::EncodeMetric + std::fmt::Debug {}
+
+impl<T> Metric for T where T: crate::encoding::EncodeMetric + std::fmt::Debug {}
--- a/src/metrics/family.rs
+++ b/src/metrics/family.rs
@@ -333,7 +333,7 @@ impl<S: EncodeLabelSet, M: EncodeMetric + TypedMetric, T: Iterator<Item = (S, M)
     fn encode(&self, mut encoder: MetricEncoder<'_, '_>) -> Result<(), std::fmt::Error> {
         let mut iter = self.borrow_mut();
 
-        while let Some((label_set, m)) = iter.next() {
+        for (label_set, m) in iter.by_ref() {
             let encoder = encoder.encode_family(&label_set)?;
             m.encode(encoder)?;
         }
--- a/src/metrics/family.rs
+++ b/src/metrics/family.rs
@@ -2,6 +2,8 @@
 //!
 //! See [`Family`] for details.
 
+use crate::encoding::{EncodeLabelSet, EncodeMetric, MetricEncoder};
+
 use super::{MetricType, TypedMetric};
 use parking_lot::{MappedRwLockReadGuard, RwLock, RwLockReadGuard, RwLockWriteGuard};
 use std::collections::HashMap;
@@ -41,23 +43,24 @@ use std::sync::Arc;
 /// family.get_or_create(&vec![("method".to_owned(), "GET".to_owned())]).inc();
 ///
 /// # // Encode all metrics in the registry in the text format.
-/// # let mut buffer = vec![];
+/// # let mut buffer = String::new();
 /// # encode(&mut buffer, &registry).unwrap();
 /// #
 /// # let expected = "# HELP my_counter This is my counter.\n".to_owned() +
 /// #                "# TYPE my_counter counter\n" +
 /// #                "my_counter_total{method=\"GET\"} 1\n" +
 /// #                "# EOF\n";
-/// # assert_eq!(expected, String::from_utf8(buffer).unwrap());
+/// # assert_eq!(expected, buffer);
 /// ```
 ///
 /// ### [`Family`] with custom type for performance and/or type safety
 ///
-/// Using `Encode` derive macro to generate
-/// [`Encode`](crate::encoding::text::Encode) implementation.
+/// Using `EncodeLabelSet` and `EncodeLabelValue` derive macro to generate
+/// [`EncodeLabelSet`] for `struct`s and
+/// [`EncodeLabelValue`](crate::encoding::EncodeLabelValue) for `enum`s.
 ///
 /// ```
-/// # use prometheus_client::encoding::Encode;
+/// # use prometheus_client::encoding::{EncodeLabelSet, EncodeLabelValue};
 /// # use prometheus_client::encoding::text::encode;
 /// # use prometheus_client::metrics::counter::{Atomic, Counter};
 /// # use prometheus_client::metrics::family::Family;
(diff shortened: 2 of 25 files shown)


src/collector.rs
```diff
@@ -0,0 +1,9 @@
+// TODO: Document
+
+use std::borrow::Cow;
+
+use crate::registry::Descriptor;
+
+pub trait Collector<M: Clone> {
+    fn collect<'a>(&'a self) -> Box<dyn Iterator<Item = (Cow<Descriptor>, Cow<M>)> + 'a>;
+}
```
mxinden, line 9: This is the trait one would need to implement to provide a custom collector, e.g. a process collector.

src/registry.rs
```diff
@@ -151,22 +164,45 @@ impl<M> Registry<M> {
...
+    ///     );
+    ///     Box::new(std::iter::once((Cow::Owned(descriptor), Cow::Owned(c))))
+    ///   }
+    /// }
+    ///
+    /// let my_collector = Box::new(MyCollector{});
+    ///
+    /// let mut registry: Registry<Counter> = Registry::default();
+    ///
+    /// registry.register_collector(my_collector);
```
mxinden, line 200: Here is an example on how to register a custom collector.

mxinden (commented):
What do folks think of this design? Would this enable your use-cases?

//CC @gagbo @dovreshef @vladvasiliu @phyber 

Alternative implementation to https://github.com/prometheus/client_rust/pull/82.

src/registry.rs
```diff
@@ -57,26 +59,37 @@ use std::borrow::Cow;
 /// #                "# EOF\n";
 /// # assert_eq!(expected, String::from_utf8(buffer).unwrap());
 /// ```
-#[derive(Debug)]
 pub struct Registry<M = Box<dyn crate::encoding::text::SendSyncEncodeMetric>> {
```
mxinden, line 62: With the `Collector` mechanism I don't think there is need for being abstract over `M` any longer. Using a concrete type (e.g. `enum` over all metrics in `prometheus_client::metrics`) will drastically simplify both the library and its usage.

mxinden:
Also //CC @sd2k

mxinden:
And //CC @08d2

dovreshef:
Looks like it will work for the process collector use case! Thanks.

vladvasiliu:
Am I understanding correctly that the user of the lib is expected to reimplement the `Collector` trait such that calling `collect()` will do whatever is necessary to produce the metrics, possibly querying some external system?

If so, I think there should be a way to provide an async collector; this will probably require implementing some alternative AsyncRegistry with matching AsyncCollector trait.

08d2:
> Am I understanding correctly that the user of the lib is expected to reimplement the `Collector` trait such that calling `collect()` will do whatever is necessary to produce the metrics, possibly querying some external system?
> 
> If so, I think there should be a way to provide an async collector; this will probably require implementing some alternative AsyncRegistry with matching AsyncCollector trait.

For context, `collect()` is called when Prometheus does an HTTP GET of the metrics endpoint for a process. That call is subject to a few layers of (typically sub-second) timeouts and failsafes and is expected to return more or less immediately. An implementation of `collect()` which blocked in any way that would benefit from async would almost certainly be a design error. If the information returned by `collect()` needed to come from an external source, then it's I guess a de facto requirement that the retrieval is decoupled from the collecting.

vladvasiliu:
That's what I was thinking, although [this example](https://github.com/prometheus/client_java/blob/master/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/StandardExports.java) seems to be reading stuff (memory info) from an external system (though, in this case, it's probably unlikely that it should take long).

I have to think a bit more about this; it seems to me that once the collector is added to the registry, there's no easy way to only update it such that it doesn't share state between possibly concurrent calls to the `/metrics` endpoint, all the while updating said state outside `collect()`.

08d2:
> although [this example](https://github.com/prometheus/client_java/blob/master/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/StandardExports.java) seems to be reading stuff (memory info) from an external system

callLongGetter? Yeah, it's not obvious to me what that does, exactly... but I wouldn't take too much inspiration from the Java client, which hasn't been updated in a long while, and was never particularly idiomatic. Stick with the Go client, or maybe look at the more active exporters hosted in the Prometheus org.

vladvasiliu:
> callLongGetter? Yeah, it's not obvious to me what that does, exactly... but I wouldn't take too much inspiration from the Java client, which hasn't been updated in a long while, and was never particularly idiomatic. Stick with the Go client, or maybe look at the more active exporters hosted in the Prometheus org.

I was thinking more about `collectMemoryMetricsLinux`. The reason I gave this example was because it's the one mentioned on the docs.

> The nature of the problem demands shared state, for sure. (Or a "clever" workaround to avoid it, which, well.) But does shared state imply async, either necessarily or even ideally? I don't think so, but I could be wrong.

I wouldn't say shared state implies async. I think the issue was the way I was thinking about somehow getting the external (non-shared) state as a collector of the shared state, which would have then required `collect()` to cover async collection.

This could probably be implemented the other way around, with the shared state added as a sub-collector to the non-shared one before returning. Which, if `collect()` is guaranteed to return quickly, could be called from an async context.

vladvasiliu:
I've been looking a bit more over this, and I think that, basically, a `Registry` is a `Collector`: it holds a bunch of metrics, right? [That's how the Python client is implemented](https://github.com/prometheus/client_python/blob/master/prometheus_client/registry.py).

I think that if the lib provided an implementation of `Collector` for `Registry`, this would solve the majority of use cases.

src/collector.rs
sd2k: I see this trait is generic over `M`, which I think is the type of the metric? How does the `Clone` bound work with collectors/registries of type `Box<dyn EncodeMetric>` which aren't `Clone`? 🤔

mxinden:
> Am I understanding correctly that the user of the lib is expected to reimplement the `Collector` trait such that calling `collect()` will do whatever is necessary to produce the metrics, possibly querying some external system?

Correct.

---



> If so, I think there should be a way to provide an async collector; this will probably require implementing some alternative AsyncRegistry with matching AsyncCollector trait.

Can you be more concrete? Can you describe a use-case where one would need an async collector?

---

> That call is subject to a few layers of (typically sub-second) timeouts and failsafes, and is expected to return more or less immediately.

Prometheus does not expect the scrape call to return immediately. E.g. from the [Prometheus scheduling section](https://prometheus.io/docs/instrumenting/writing_exporters/#scheduling):

> The default scrape timeout for Prometheus is 10 seconds.

---

> If the information returned by `collect()` needed to come from an external source, then it's I guess a de facto requirement that the retrieval is decoupled from the collecting.

By default the collection should not be decoupled from the collection. Citing the [Prometheus scheduling section](https://prometheus.io/docs/instrumenting/writing_exporters/#scheduling):

> Metrics should only be pulled from the application when Prometheus scrapes them, exporters should not perform scrapes based on their own timers. That is, all scrapes should be synchronous.

---

> I think that if the lib provided an implementation of `Collector` for `Registry`, this would solve the majority of use cases.

`Registry` could implement `Collector`, though I have refrained from that thus far to reduce complexity, i.e. not to mix concepts. I expect a user to have a single `Registry` but many `Collectors.

I don't follow how `Registry` implementing `Collector` would solve an issue here. Would you mind expanding on that?

mxinden:
Also note that the `encode` function take an immutable reference to a `Registry`, thus two Prometheus servers (e.g. for fault tolerance) may scrape the same endpoint concurrently.

src/collector.rs
mxinden: > How does the `Clone` bound work with collectors/registries of type `Box<dyn EncodeMetric>` which aren't `Clone`? thinking

It doesn't. For now this is just a proposal. We could require `Clone` on `EnocdeMetric`, though I think the much better idea is to no longer be generic over the metric type. See https://github.com/prometheus/client_rust/pull/82#discussion_r956892586.

Let me know what you think @sd2k!

sd2k: Ah great, that makes sense. I agree that switching to an enum instead of generics/trait objects would simplify things a lot! (Especially since adding a `Clone` bound to `EncodeMetric` would mean it wasn't object safe, so I don't think we could use trait objects anyway).

mxinden: > (Especially since adding a Clone bound to EncodeMetric would mean it wasn't object safe, so I don't think we could use trait objects anyway).

Good point. We would have to drop the `Cow<M>` in favor of `M`, but then again, let's try to get rid of `M`.

vladvasiliu:
> Can you be more concrete? Can you describe a use-case where one would need an async collector?

My use case is exporting metrics that are gathered via multiple HTTP calls. If my HTTP client is async, `collect()` should be able to handle that.

> Prometheus does not expect the scrape call to return immediately. E.g. from the [Prometheus scheduling section](https://prometheus.io/docs/instrumenting/writing_exporters/#scheduling):

Well, the question is whether `collect()` itself should return immediately. The other operations could be done outside it, while staying inside the global scrape.

> I don't follow how Registry implementing Collector would solve an issue here. Would you mind expanding on that?

See the other point: this is a confusion about whether `collect()` should be guaranteed to be quick or not and whether a collector should be different from a registry.

If it's not, and random long operations can be done inside `collect()`, then I can see how a `Collector` can be different from a `Registry`, and thus providing an default implementation for Registry doesn't help, since every collector is expected to be different and do its own thing.

However, in the second situation, when the actual gathering of the metrics should be done outside of the `collect()` method, then basically this becomes generating two separate registries, and combining them before calling `collect()`. The idea wouldn't be to do the scraping outside of the call from prometheus, but only outside the call to `collect()`.

08d2:
That example describes how a collector might be "bolted on" to a legacy system. It's not indicative of  best practice. But the point is largely moot, I think.

kwiesmueller:
My current examples why a collector would need to support async are:
* collecting metrics from devices via e.g. Bluetooth
* collecting container metrics comparable to cadvisor where the list of metrics is basically generated from a current state that is retrieved from the filesystem, containerd APIs etc.

While both approaches could also refresh the data in the background, having it directly connected to the scrape and being able to fetch fresh data on a scrape provides more reliable resolution than scraping previously scraped data.

Now, the data refresh could actually be handled in the http handler, but still it seems like an unnecessary detour.

mxinden:
As a first iteration, I suggest we support synchronous `Collector::collect` only.

1. For those that need to call `async` methods in their `Collector::collect` call, I suggest you spawn the `encode` call in a new OS thread (or in a blocking-aware runtime task) and call any `async` methods via [`block_on`](https://docs.rs/futures/latest/futures/executor/fn.block_on.html) in your `Collector::collect` implementation.
2. For those that need concurrency within your `Collector::collect` implementation (e.g. reaching out to two remote machines concurrently) I suggest you spawn two or more OS threads within your `Collector::collect` implementation and join the threads again.
3. For those that need the `Collector::collect` calls of two distinct `Collector` implementations to run concurrently, you would be out of luck **for now.**

In case folks see a large performance hit, e.g. due to thread spawns (even though that should be < 10 microseconds), or (1) and (2) add too much complexity in your `Collector::collect` implementation or folks need (3) I suggest we design an `async` `Collector::collect` as a next step.

mxinden:
@kwiesmueller Thanks for sharing. Let me know in case the threading approach is good enough for you for now, i.e. can collect from the many datasources you are targeting within the Prometheus scrape timeout.

If not, as mentioned above, let's design an `async` `Collector::collect` as a next step.

mxinden:
I rebased this pull request onto https://github.com/prometheus/client_rust/pull/100. Everything compiles, tests succeed.

To me the above validates that we should move forward with https://github.com/prometheus/client_rust/pull/100. That said, I do think the signatures in this pull request need more work (e.g. the workaround with `MaybeOwned`).

src/registry.rs
mxinden: See https://github.com/prometheus/client_rust/pull/100 proposing the above.

hexfusion:
Very interested in this work are you looking for help or what would help move this forward.

mxinden:
> Very interested in this work 

Thanks for the interest. Sorry for the silence.

I rebased this pull request on https://github.com/prometheus/client_rust/pull/105. I want to try out the new API with one of my projects (e.g. https://github.com/mxinden/kademlia-exporter/) before merging #105 and this pull request.

> are you looking for help or what would help move this forward.

I would appreciate feedback on #105 and this pull request. You would be of great help by testing this pull request (it is on top of #105) within your application in need of the `Collector` abstraction @hexfusion.

hexfusion:
I have some cycles today will take a look, thanks!

mxinden:
Tested this patch with `rust-libp2p` and `kademlia-exporter`. While the `Collector` trait is still quite verbose (with `Cow` and `MaybeOwned`), it does function as expected and allows ad-hoc cheap metric generation.

https://github.com/mxinden/kademlia-exporter/pull/209

src/registry.rs
```diff
@@ -228,41 +258,58 @@ impl Registry {
             .expect("sub_registries not to be empty.")
     }
 
-    /// [`Iterator`] over all metrics registered with the [`Registry`].
-    pub fn iter(&self) -> RegistryIterator {
```
thomaseizinger, line 232: Renaming `RegistryIterator` and making this `pub(crate)` is the only breaking change as far as I can tell. Might be worth undoing.

src/metrics/family.rs
```diff
@@ -326,6 +327,24 @@ where
     }
 }
 
+impl<S: EncodeLabelSet, M: EncodeMetric + TypedMetric, T: Iterator<Item = (S, M)>> EncodeMetric
+    for RefCell<T>
```
thomaseizinger, line 331: This is very hard to discover and doesn't feel Rust-y to me. Why do we need to depend on `RefCell` here?

I think it would be more idiomatic to do:

- `impl EncodeMetric for Vec<T> where T: EncodeMetric { }`
- `impl EncodeMetric for (S, M) where S: EncodeLabelSet, M: EncodeMetric { }`

Then, have the user provide a `Vec` instead of being generic over something that can be iterated. That should get rid of the `RefCell` and iterator dependency and would compose better with other usecases.

mxinden: An `Iterator::next` takes `&mut self` in order to be able to e.g. track its position within a `Vec`. Unfortunately `EncodeMetric::encode` takes `&self`.

In this pull request `RefCell` is used for interior mutability. It allows calling an `Iterator::next` taking `&mut self` from an `EncodeMetric::encode` providing `&self` only.

An alternative approach would be to change `EncodeMetric::encode` to take `&mut self`. I deemed the change and its implications not worth it, given that `Collector` is an abstraction for advanced users only.

All that said, I agree that the usage of `RefCell` is neither simple nor intuitive. @thomaseizinger do you see other alternatives to the one above?

PR #84, opened 2022-08-30
src/lib: Add `#![deny(missing_docs)]`

mxinden:
Enforces documentation comments on every public type.


phyber (approved):
Looks like it does what it says on the tin. New docs seem fine.

PR #85, opened 2022-08-31
src/metrics/family.rs: Add remove and clear methods, and tests

phyber:
This PR adds `Family::remove` as discussed a little in #36 along with a basic test for the behaviour. We also add a `Family::clear` method as discussed in this PR.

The `Family::remove` method returns a `bool` to indicate if a label set was removed, which matches what the Go client does when removing its equivalent of a label set.

The `Family::clear` method does not return anything and always successfully clears the label sets.

### DIFF
--- a/src/metrics/family.rs
+++ b/src/metrics/family.rs
@@ -3,7 +3,7 @@
 //! See [`Family`] for details.
 
 use super::{MetricType, TypedMetric};
-use parking_lot::{MappedRwLockReadGuard, RwLock, RwLockReadGuard};
+use parking_lot::{MappedRwLockReadGuard, RwLock, RwLockReadGuard, RwLockWriteGuard};
 use std::collections::HashMap;
 use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


src/metrics/family.rs
```diff
@@ -234,9 +234,35 @@ impl<S: Clone + std::hash::Hash + Eq, M, C: MetricConstructor<M>> Family<S, M, C
...
+        self.write().remove(label_set).is_some()
+    }
+
     pub(crate) fn read(&self) -> RwLockReadGuard<HashMap<S, M>> {
         self.metrics.read()
     }
+
+    fn write(&self) -> RwLockWriteGuard<HashMap<S, M>> {
+        self.metrics.write()
+    }
```
mxinden, line 265: I am guessing you added this method for the sake of consistency with the `read` method above? I don't think it is necessary. The `read` method only exists as code within the `crate` but outside this `mod` needs it.

src/metrics/family.rs
```diff
@@ -234,9 +234,35 @@ impl<S: Clone + std::hash::Hash + Eq, M, C: MetricConstructor<M>> Family<S, M, C
...
+    ///
+    /// // Will create the metric with label `method="GET"` on first call and
+    /// // return a reference.
+    /// family.get_or_create(&vec![("method".to_owned(), "GET".to_owned())]).inc();
+    ///
+    /// // Will return `true`, indicating that the `method="GET"` label set was
+    /// // removed.
+    /// assert!(family.remove_label_set(&vec![("method".to_owned(), "GET".to_owned())]));
+    /// ```
+    pub fn remove_label_set(&self, label_set: &S) -> bool {
```
mxinden, line 255:
```suggestion
    pub fn remove(&self, label_set: &S) -> bool {
```

What do you think of being consistent with the other implementations? Do you foresee the need for any other `remove_*` styles in the future?

mxinden (commented):
Solid work! Thanks, especially for the doc comment and example!

According to the client library documentation there should be both a `remove` and a `clear`. I don't have a strong opinion for the latter. But given that is should be easy to add, what do you think of including it in this pull request?

> Metrics with labels SHOULD support a remove() method with the same signature as labels() that will remove a Child from the metric no longer exporting it, and a clear() method that removes all Children from the metric. These invalidate caching of Children.

https://prometheus.io/docs/instrumenting/writing_clientlibs/#labels

src/metrics/family.rs
phyber: Yeah. This was for consistency. Im happy to remove it if needed.

src/metrics/family.rs
phyber: I think I'd be happy to rename this to simply `remove`.

phyber:
I'd be happy to work on adding a `clear` method here as well. I might not have that before the end of this week though, as I'm away for a few days :)

mxinden:
No rush @phyber. Thanks for the help.

phyber:
`Family::remove_label_sets` has now been renamed to `Family::remove`, and a `Family::clear` method has been added.

The redundant `Family::write` private method was removed.

mxinden (commented):
Thanks for the follow-ups!

Just needs a bump of the patch version in `Cargo.toml` and a changelog entry in `CHANGELOG.md`.

phyber:
Rebased onto the current `master` and added entries to the CHANGELOG for this PR.

I didn’t bump the version in `Cargo.toml` since I guess this will all be released under the (currently unreleased) v0.19.0.

mxinden (approved):
:pray:

PR #93, opened 2022-09-17
derive-encode: Fix `multiple applicable items in scope` error

ackintosh:
`multiple applicable items in scope` error happens when there is an `encode` implementation (e.g. [`prost::Message`](https://docs.rs/prost/latest/prost/trait.Message.html)) in the same scope.

```shell
        // error[E0034]: multiple applicable items in scope
        //     |
        // 120 |         #[derive(Clone, Hash, PartialEq, Eq, Encode)]
        //     |                                              ^^^^^^ multiple `encode` found
```

### DIFF
--- a/derive-encode/src/lib.rs
+++ b/derive-encode/src/lib.rs
@@ -104,7 +104,7 @@ fn derive_protobuf_encode(ast: DeriveInput) -> TokenStream2 {
                         quote! {
                             let mut label = {
                                 let mut labels = vec![];
-                                self.#ident.encode(&mut labels);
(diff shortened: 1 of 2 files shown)


ackintosh:
Okay, the unit test has failed as expected.

https://github.com/prometheus/client_rust/actions/runs/3072132821/jobs/4963408836

ackintosh:
I think a change log entry for this PR is not needed as v0.19 (protobuf feature) is not released. Let me know if it's needed. 🙂

mxinden (approved):
Looks good to me. Thanks for the fix. No changelog entry needed.

Sorry for the delay on releasing this.

Issue #25, opened 2021-11-12
Global registry pattern


thedodd:
For systems where there are many components which need to register their own metrics, shared mutable access to the registry is required in order to keep things nice and simple (for the user). Providing a built-in solution for this pattern in the crate would lower the barrier of adoption, and could also help folks not to mess up their own implementation of the pattern where they might code the mutex incorrectly.

The prometheus crate has such a pattern, and it definitely makes getting started quite easy.

mxinden:
I am undecided on the topic of global registries. Not offering it within `open-metrics-client` thus far has been by design.

On the one hand I do understand that offering a global metrics registry within `open-metrics-client` directly simplifies instrumentation across ones application.

On the other hand:

- I consider global variables in general and a global metric registry in specific error prone. The implicit side effect of registering metrics with a global registry might catch folks by surprise, e.g. when importing an external library for the sake of a single helper function.
- Instead of `open-metrics-client` offering a global registry, users can define their own within their binary.

My proposition moving forward: Instead of adding a global registry to `open-metrics-client`, add an example showcasing how users can define it themselves, including a disclaimer on how they might be shooting themselves in the foot.

adamchalmers:
Hi! I think making an example of global registries would be a great solution. I'd love to switch over to this library but I'm a bit reluctant to sit down and do the work of figuring out a global registry myself. Has anyone done this yet? I'd happily contribute examples or even an explainer post on my Rust programming blog. I just haven't sat down yet to figure out how to implement a global registry.

huntc:
From my reading of the client library doco, there appears to be a requirement for a global registry (default collector registry): https://prometheus.io/docs/instrumenting/writing_clientlibs/#metrics

I would suppose that having a global registry is no different to having a global memory  allocator.

Taking this further, having static/compile time registration could also make sense. The client library doco also appears to encourage this in the same section referenced above, albeit static, not const.

adamchalmers:
I actually think this is a non-issue. You don't need a global registry, you need global _metrics_. Your individual metrics can be global, and accessible by both your registry and your normal code. The registry can read them and serve them without being global! 

See my example [here](https://github.com/prometheus/client_rust/pull/94/files#diff-7430344b39ea433c99f1cf84a5cc3118954fae4afaeb09282e69f34c538860db). Here I have a local registry, which is owned by the HTTP server which serves metrics. But my counter is in a `lazy_static` block, and can be read by both the registry and my normal business logic code.

[...]

PR #94, opened 2022-09-19
Example metrics server using Hyper

adamchalmers:
Say your Rust app is not a web server. You probably don't need to pull in a whole web framework like Axum or Actix-web just to serve one little OpenMetrics endpoint. You can just use Hyper instead. Here's an example of how to do that.

I also added this an example target in Cargo.toml, so that you can run it with cargo:

```sh
$ cargo run --example hyper  &
$ curl localhost:8001
# HELP tokio_hyper_example_example_counter An example counter that will get incremented automatically.
# TYPE tokio_hyper_example_example_counter counter
tokio_hyper_example_example_counter_total 6
# EOF
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,8 @@ quickcheck = "1"
 rand = "0.8.4"
 tide = "0.16"
 actix-web = "4"
+tokio = { version = "1", features = ["rt-multi-thread", "net", "macros", "signal"] }
+hyper = { version = "0.14.16", features = ["server", "http1", "tcp"] }
 
 [build-dependencies]
 prost-build = { version = "0.9.0", optional = true }
--- a/examples/hyper.rs
+++ b/examples/hyper.rs
@@ -1,0 +1,75 @@
+use hyper::{
+    service::{make_service_fn, service_fn},
+    Body, Request, Response, Server,
+};
+use prometheus_client::{encoding::text::encode, metrics::counter::Counter, registry::Registry};
+use std::{
+    future::Future,
+    io,
+    net::{IpAddr, Ipv4Addr, SocketAddr},
+    pin::Pin,
+    sync::Arc,
+};
+use tokio::signal::unix::{signal, SignalKind};
+
+#[tokio::main]
+async fn main() {
+    let request_counter: Counter<u64> = Default::default();
+
+    let mut registry = <Registry>::with_prefix("tokio_hyper_example");
+
+    registry.register(
+        "requests",
+        "How many requests the application has received",
+        Box::new(request_counter.clone()),
+    );
+
+    // Spawn a server to serve the OpenMetrics endpoint.
+    let metrics_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8001);
+    start_metrics_server(metrics_addr, registry).await
+}
+
+/// Start a HTTP server to report metrics.
+pub async fn start_metrics_server(metrics_addr: SocketAddr, registry: Registry) {
+    let mut shutdown_stream = signal(SignalKind::terminate()).unwrap();
+
+    eprintln!("Starting metrics server on {metrics_addr}");
+
+    let registry = Arc::new(registry);
+    Server::bind(&metrics_addr)
+        .serve(make_service_fn(move |_conn| {
+            let registry = registry.clone();
+            async move {
+                let handler = make_handler(registry);
+                Ok::<_, io::Error>(service_fn(handler))
+            }
+        }))
+        .with_graceful_shutdown(async move {
+            shutdown_stream.recv().await;
+        })
+        .await
+        .unwrap();
+}
+
+/// This function returns a HTTP handler (i.e. another function)
+pub fn make_handler(
+    registry: Arc<Registry>,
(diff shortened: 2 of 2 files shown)


examples/hyper.rs
```diff
@@ -0,0 +1,89 @@
...
+    move |_req: Request<Body>| {
+        let reg = registry.clone();
+        Box::pin(async move {
+            let mut buf = Vec::new();
+            encode(&mut buf, &reg.clone()).map(|_| {
+                let body = std::str::from_utf8(buf.as_slice()).unwrap().to_string();
+                Response::builder()
+                    .header(
+                        hyper::header::CONTENT_TYPE,
+                        "application/openmetrics-text; version=1.0.0; charset=utf-8",
```
mxinden, line 82: :pray:

examples/hyper.rs
```diff
@@ -0,0 +1,89 @@
...
+    net::{IpAddr, Ipv4Addr, SocketAddr},
+    pin::Pin,
+    sync::Arc,
+    time::Duration,
+};
+use tokio::signal::unix::{signal, SignalKind};
+
+lazy_static::lazy_static! {
+    pub static ref EXAMPLE_COUNTER: Counter = Counter::default();
+}
```
mxinden, line 18: I would like to encourage users to use local metrics, thus I would prefer this example to use a locally instantiated metric.

Would you mind changing this example to use a non-global metric instead?

I am fine with having an example using global metrics in this repository, though I would like that to be a dedicated example, documenting the trade offs a user buys into.

See also https://github.com/prometheus/client_rust/issues/25

examples/hyper.rs
```diff
@@ -0,0 +1,89 @@
...
+    let registry = register_metrics();
+
+    // Spawn a server to serve the OpenMetrics endpoint.
+    let metrics_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8001);
+    tokio::task::spawn(start_server(metrics_addr, registry));
+
+    loop {
+        EXAMPLE_COUNTER.inc();
+        tokio::time::sleep(Duration::from_secs(1)).await;
+    }
```
mxinden, line 31: This is a neat idea, though for the sake of consistency with the other examples, I would prefer showcasing a standard HTTP request counter in this example.

See tide example:

https://github.com/prometheus/client_rust/blob/12c1de56116d803c709ed6584c263353727221fa/examples/tide.rs#L47-L57

I think an example should showcase a single thing only. (Though it is hard where the boundaries of a thing are.) I am fine for something like the above to be shown in a separate example.

Cargo.toml
```diff
@@ -34,10 +34,17 @@ quickcheck = "1"
...
+hyper = { version = "0.14.16", features = ["server", "http1", "tcp"] }
+lazy_static = "1.4"
 
 [build-dependencies]
 prost-build = { version = "0.9.0", optional = true }
 
+[[example]]
+name = "hyper"
+crate-type = ["bin"]
+
```
mxinden, line 47: Sorry for the potentialy stupid question: Why is this entry needed? I can already do `cargo run --example tide` today. Wouldn't `cargo run --example hyper` work just fine without the lines above?

mxinden (commented):
:rocket: Thanks for the help. Very much appreciated.

Cargo.toml
adamchalmers: You're right, it's not needed, I mistakenly thought it was.

examples/hyper.rs
```diff
@@ -0,0 +1,81 @@
...
+pub fn make_handler(
+    registry: Arc<Registry>,
+) -> impl Fn(Request<Body>) -> Pin<Box<dyn Future<Output = io::Result<Response<Body>>> + Send>> {
+    // This closure accepts a request and responds with the OpenMetrics encoding of our metrics.
+    move |_req: Request<Body>| {
+        let reg = registry.clone();
+        Box::pin(async move {
+            let mut buf = Vec::new();
+            encode(&mut buf, &reg.clone()).map(|_| {
+                let body = std::str::from_utf8(buf.as_slice()).unwrap().to_string();
```
mxinden, line 70: Given that one can build a `Body` from a `[u8]` why is the conversion to a string necessary here?

https://docs.rs/hyper/latest/hyper/struct.Body.html#impl-From%3C%26%27static%20%5Bu8%5D%3E-for-Body

mxinden (commented):
Just two comments. Otherwise looks good to me. Thanks for the follow-ups.

examples/hyper.rs
```diff
@@ -0,0 +1,81 @@
...
+    pin::Pin,
+    sync::Arc,
+};
+use tokio::signal::unix::{signal, SignalKind};
+
+#[tokio::main]
+async fn main() {
+    let request_counter = Counter::default();
+
+    let registry = register_metrics(&request_counter);
```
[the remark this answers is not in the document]
adamchalmers: I removed the function, you're right

examples/hyper.rs
adamchalmers: Ah great point, thanks for pointing that out. Simplified it.

mxinden (approved):
Thanks for the help. Thanks for making it easier for others to get started.

PR #83, opened 2022-08-29
Protobuf support

ackintosh:
Closes #46 

This PR adds protobuf support to the client, which provides functionality to encode metrics to an instance of the [OpenMetrics protobuf](https://github.com/OpenObservability/OpenMetrics/blob/main/proto/openmetrics_data_model.proto) format.

```rust
        let mut registry = Registry::default();

        // ...

        // Encode metrics to an instance of the OpenMetrics protobuf format.
        let metric_set: openmetrics_data_model::MetricSet = encode(&registry);

        for family in metric_set.metric_families {
            println!(family.name);

            for metric in family.metrics {
                // ...
            }
        }
```

PR #96, opened 2022-09-19
impl TypedMetric for Counter/HistogramWithExemplars

### LINKED ISSUE #95: Bug: cannot use Exemplars and Families together
Hi all, thanks for making this library, I'm really excited to get started with exemplars. But I can't see how to combine exemplars and families.

**Problem**
By my read of the OpenMetrics spec, I should be able to export metrics like this:
```
latency_bucket{result="success",le="0.00256"} 27300 # {trace_id="3a2f90c9f80b894f"} 0.001345422
```
Note that the `latency` metric has labels in both its family (`result`) and its exemplars (`trace_id`)

I am trying to implement that using this library, but it won't compile. Why? Because the trait `TypedMetric` is not implemented for `CounterWithExemplar<ExemplarLabel>` and that means my metric does not impl `SendSyncEncodeMetric`. So it can't be registered. See this example:

```rust
use prometheus_client::{
    encoding::text::SendSyncEncodeMetric,
    metrics::{exemplar::HistogramWithExemplars, family::Family, histogram::exponential_buckets},
    registry::Registry,
};
use prometheus_client_derive_encode::Encode;

fn main() {
    // Create a metric registry.

    // Metric will be used something like this:
    // latency_bucket{result="success",le="0.00256"} 27300 # {trace_id="3a2f90c9f80b894f"} 0.001345422 1.6188203823429482e+09
    let latency: Family<ResultLabel, HistogramWithExemplars<TraceLabel>> =
        Family::new_with_constructor(|| {
            HistogramWithExemplars::new(exponential_buckets(1.0, 2.0, 10))
        });

    // This line doesn't compile:
    //
    // the trait `TypedMetric` is not implemented for `HistogramWithExemplar<TraceLabel>`
    let b: Box<dyn SendSyncEncodeMetric> = Box::new(latency);

    // Register metrics.
    registry.register("latency", "help text goes here", b);
}

#[derive(Clone, Hash, PartialEq, Eq, Encode, Debug, Default)]
pub struct ResultLabel {
    pub result: String,
}

#[derive(Clone, Hash, PartialEq, Eq, Encode, Debug, Default)]
pub struct TraceLabel {
    trace_id: String,
}
```

adamchalmers:
Fixes #95, i.e. that on master, you cannot create a Histogram or Counter which has both a Family and Exemplars. Also adds an example showing how to make such a histogram/counter. This example won't compile without my fix, so it also serves as a regression test :)

Running my example gives the expected output:
```
# HELP latency help text goes here.
# TYPE latency histogram
latency_sum{result="success"} 0.001345422
latency_count{result="success"} 1
latency_bucket{result="success",le="1.0"} 1 # {trace_id="3a2f90c9f80b894f"} 0.001345422
```

If you approve this PR, would you please also consider cutting a release 0.19 for it? I think most real-world exemplars will also want to use families, so I think this feature is worth releasing soon. My team at Cloudflare is gonna use my fork with this PR for now, but I'd like to start using this crate in more projects internally, and we try to avoid relying on Github forks :)

Thanks again for making this library. I've been using the old prometheus crate since 2018, and made a few PRs here and there, so I really appreciate you modernizing it and adding full OpenMetrics compliance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   set from a family. See [PR 85].
 - Added a `clear` method to `Family` to allow the removal of all label sets
   from a family. See [PR 85].
+- Impl `TypedMetric` for `CounterWithExemplar` and `HistogramWithExemplar`, so that they can be used with `Family`. See [PR 96].
 
 ### Changed
 
@@ -19,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 [PR 83]: https://github.com/prometheus/client_rust/pull/83
 [PR 85]: https://github.com/prometheus/client_rust/pull/85
+[PR 96]: https://github.com/prometheus/client_rust/pull/96
 
 ## [0.18.0]
 
--- a/src/metrics/exemplar.rs
+++ b/src/metrics/exemplar.rs
@@ -4,6 +4,7 @@
 
 use super::counter::{self, Counter};
 use super::histogram::Histogram;
+use super::{MetricType, TypedMetric};
 use parking_lot::{MappedRwLockReadGuard, RwLock, RwLockReadGuard};
 use std::collections::HashMap;
 #[cfg(any(target_arch = "mips", target_arch = "powerpc"))]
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   set from a family. See [PR 85].
 - Added a `clear` method to `Family` to allow the removal of all label sets
   from a family. See [PR 85].
+- Impl `TypedMetric` for `CounterWithExemplar` and `HistogramWithExemplar`, so that they can be used with `Family`. See [PR 96].
```
bobrik, line 15: You need to add a link to `PR 96`, search for `PR 95` below.

bobrik (approved):

examples/families-exemplars.rs
```diff
@@ -0,0 +1,47 @@
+use prometheus_client::{
```
mxinden, line 1: Thanks for also adding an example. I wonder whether people would easier discover this as a doc example on `HistogramWithExemplars` or `Family` (potentially one linking to the other). What do you think @adamchalmers?

Cargo.toml
```diff
@@ -38,6 +38,9 @@ actix-web = "4"
 [build-dependencies]
 prost-build = { version = "0.9.0", optional = true }
 
+[[example]]
```
mxinden, line 41: Cross-referencing question on whether this is needed: https://github.com/prometheus/client_rust/pull/94#discussion_r976458080

mxinden (commented):
Oh, my bad. Thanks for the fix. Much appreciated.

mxinden:
> If you approve this PR, would you please also consider cutting a release 0.19 for it? I think most real-world exemplars will also want to use families, so I think this feature is worth releasing soon. My team at Cloudflare is gonna use my fork with this PR for now, but I'd like to start using this crate in more projects internally, and we try to avoid relying on Github forks :)

:+1:  Yes, happy to cut a new release once this is merged, especially since that would publish protobuf support https://github.com/prometheus/client_rust/pull/83.

> Thanks again for making this library. I've been using the old prometheus crate since 2018, and made a few PRs here and there, so I really appreciate you modernizing it and adding full OpenMetrics compliance.

Wonderful feedback. Thank you :heart:

examples/families-exemplars.rs
adamchalmers: Sounds good, I'll remove this example and replace it with examples on the docs for CounterWithExemplar and HistogramWithExemplars.

Cargo.toml
adamchalmers: Ah, it's not needed! I don't know why I thought it was. Removed anyway in favour of using examples on the docs.

adamchalmers:
Hmm, given that you're shipping protobuf support, you may want to include #98 in the next release too.

src/metrics/exemplar.rs
```diff
@@ -31,12 +32,46 @@ pub struct Exemplar<S, V> {
...
+/// latency
+/// .get_or_create(&ResultLabel {
+///     result: "success".to_owned(),
+/// })
+/// .inc_by(
+///     1,
+///     Some(TraceLabel {
+///         trace_id: "3a2f90c9f80b894f".to_owned(),
+///     }),
+/// );
```
mxinden, line 63: Can you fix the indentation here? As far as I know this will propagate into the rendered HTML on the docs.

src/metrics/exemplar.rs
```diff
@@ -122,12 +157,48 @@ impl<S, N: Clone, A: counter::Atomic<N>> CounterWithExemplar<S, N, A> {
...
+/// latency
+/// .get_or_create(&ResultLabel {
+///     result: "success".to_owned(),
+/// })
+/// .observe(
+///     0.001345422,
+///     Some(TraceLabel {
+///         trace_id: "3a2f90c9f80b894f".to_owned(),
+///     }),
+/// );
```
mxinden, line 190: Same here.

mxinden (commented):
Thanks for the follow-ups. Two more comments. Otherwise ready to merge.

mxinden (approved):
:pray:

PR #98, opened 2022-09-21
chore: Bump Prost to 0.11

adamchalmers:
I noticed the Dependabot PRs (#89 #90 #92), but you really have to update all three Prost crates simultaneously because otherwise you'll get weird errors between versions. So, here's one PR to bump prost's family of crates.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -7,6 +7,8 @@ jobs:
     name: Check
     runs-on: ubuntu-latest
     steps:
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
@@ -45,6 +47,8 @@ jobs:
     name: Test Suite
     runs-on: ubuntu-latest
     steps:
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
@@ -79,6 +83,8 @@ jobs:
     name: Rustfmt
     runs-on: ubuntu-latest
     steps:
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
@@ -95,6 +101,8 @@ jobs:
     name: Clippy
     runs-on: ubuntu-latest
     steps:
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
@@ -124,6 +132,8 @@ jobs:
     name: Check rustdoc intra-doc links
     runs-on: ubuntu-latest
     steps:
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
@@ -159,4 +169,4 @@ jobs:
         with:
           use-cross: true
           command: build
-          args: --release --target=${{ matrix.target }} --all-features
+          args: --release --target=${{ matrix.target }}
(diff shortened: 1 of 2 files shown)


phyber (approved):
Looks good to me.

adamchalmers:
> Looks good to me.

Thanks! As a first-time contributor, I can't merge yet, I'll need someone to allow CI jobs to run on this PR.

adamchalmers:
Can do, let me see...

adamchalmers:
@mxinden Can you pls grant me permission to run workflows? "First-time contributors need a maintainer to approve running workflows. [Learn more.](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)"

mxinden:
Approved on the latest changes. Is there a way to auto-approve your pull requests to run on CI? If not, I think you can just run the pipelines on your fork, right?

adamchalmers:
Yeah, I can run them on my fork.

adamchalmers:
I think the root cause is that the GH action for installing protoc doesn't support ARM. There's an issue open for ARM support: https://github.com/arduino/setup-protoc/issues/25 -- notice that in the [logs](https://github.com/prometheus/client_rust/actions/runs/3066919882/jobs/5033576293#step:2) from this CI job, you can see it's downloading an x86-64 binary instead of an ARM binary.

```
Downloading archive: https://github.com/protocolbuffers/protobuf/releases/download/v3.20.2/protoc-3.20.2-linux-x86_64.zip
```

I think the [erroneous line of their action is here](https://github.com/arduino/setup-protoc/blob/48d72781d31992e7902fc25136c606312e8e2454/lib/installer.js#L118):
```typescript
const arch: string = osArch == "x64" ? "x86_64" : "x86_32";
```

So we'll need a PR to that repo to add ARM support, or fork it, or write our own action to install the arm target.

mxinden:
Thanks @adamchalmers for your work on https://github.com/arduino/setup-protoc/issues/25!

adamchalmers:
My pleasure. I am a bit confused why we even need this action though -- can't we just install protoc from the package managers on windows/macos/ubuntu? Why do we need to install a particular version from GitHub release?

adamchalmers:
OK, I opened a PR to fix the issue on arduino/setup-protoc. In the meantime, I switched to my fork of the action on this branch, so we can see if CI passes. @mxinden could you please approve running CI on this branch?

mxinden:
@adamchalmers I triggered CI.

adamchalmers:
So, the reason my PR to setup-protoc doesn't work: each of this project's 5 matrix jobs (named "Cross compile") are all actually running on Ubuntu x86_64. The various CPU architectures are only passed to the `cargo build` step, they don't actually affect the action runner. So the "install protoc" step correctly recognizes the environment as Linux x86_64. 

I guess the solution is would have three parts:
1. setup-protoc needs another configurable input for target OS arch
2. The "Cross compile" job sets that input to `${{ matrix.target }}`
3. Ensure that protoc gets used for the cross-compile

mxinden:
Thanks for all the investigation @adamchalmers.

> Is it really important to test protocol buffer support on those architectures?

I think it is a nice-to-have. In other words, I am fine with not testing the `protobuf` feature on these platforms for now.

adamchalmers:
No worries! I'm a little disappointed I couldn't get it all to work how I wanted, but at least I learned a lot about GitHub actions along the way. Finally, this PR passes all the CI.

.github/workflows/rust.yml
```diff
@@ -159,4 +169,4 @@ jobs:
         with:
           use-cross: true
           command: build
-          args: --release --target=${{ matrix.target }} --all-features
+          args: --release --target=${{ matrix.target }}
```
mxinden, line 172:
```suggestion
          # Note that this does not test the `protobuf` feature (for now). See reasoning in https://github.com/prometheus/client_rust/pull/98/.
          args: --release --target=${{ matrix.target }}
```

mxinden (approved):
Again, thank you for your help here @adamchalmers.

PR #99, opened 2022-09-25
encoding/proto: Add a trait combining EncodeMetric, Send and Sync

ackintosh:
The same as [encoding::text](https://github.com/prometheus/client_rust/blob/master/src/encoding/text.rs#L406), I added `SendSyncEncodeMetric` trait combining EncodeMetric, Send, and Sync.

### DIFF
--- a/src/encoding/proto.rs
+++ b/src/encoding/proto.rs
@@ -214,6 +214,25 @@ where
     exemplar_proto
 }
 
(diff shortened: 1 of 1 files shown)


adamchalmers:
Why add another trait with the same definition as the encoding::text one? Maybe there should just be one trait that gets used in both places (text and protobuf encodings).

mxinden:
With https://github.com/prometheus/client_rust/pull/105 merged and released, this is resolved.

Issue #97, opened 2022-09-20
Consider removing text::Encode in favour of serde::Serialize


nox:
Serde's Serialize trait is implemented for dozens of types in the Rust ecosystem, its derive code is fairly well optimised, and it is the most popular serialization framework we have for Rust.

What do you think of removing the Encode trait in favour of Serialize, to get all the nice things there are in Serde for free?

mxinden:
I would be in favor of that. I should have thought of it before.

@nox would you like to write a first proposal? Before you go all the way, I would suggest writing a small proof-of-concept first to make sure using serde is a viable path forward.

nox:
Sure, I'll write a `struct Bridge<T>(T)` that implements `Encode` using `T`'s implementation of `Serialize`, in a separate crate.

nox:
I pushed https://github.com/nox/serde_prometheus_labels earlier today.

nox:
I published [serde_prometheus_labels](https://crates.io/crates/serde_prometheus_labels) which provides its own `Family<Labels, Metric>` type using serde as the encoder.

nox:
> I mean, `()` also implements Serialize/Deserialize but you probably shouldn't be able to use `()` as a label, for instance.

Why not? `()` should mean "no labels at all".

nox:
> So I'm not sure `Encode` can be entirely replaced by serde's traits, as serde's traits are basically implemented for too many types that aren't suitable.

I agree to this, but I would rather rely on serde and not be blocked on some missing implementation. It made the job for us, that's why I published that stuff as a separate crate.

mxinden:
With https://github.com/prometheus/client_rust/pull/105 I don't think we should proceed here, i.e. I don't see the value of replacing the new traits.

@nox I am closing here for now. Please comment here in case I am missing something.

Though I do believe implementing `EncodeLabelSet`, `EncodeLabel`, `EncodeLabelKey` and `EncodeLabelValue` for more types in `std` would be helpful.

PR #100, opened 2022-09-29
src/registry: Use dynamic dispatch and remove type parameter M

mxinden:
Instead of being generic over the metric type on `Registry`, use dynamic dispatch.

On the upside this significantly simplifies type signatures exposed to the user. On the downside this requires users to always use dynamic dispatch and requires all metrics to implement auxiliary trait `Debug` and auto traits `Send` and `Sync`.

Initially suggested in https://github.com/prometheus/client_rust/pull/82#discussion_r956892586

Would replace https://github.com/prometheus/client_rust/pull/99 //CC @ackintosh @adamchalmers 

Happy for alternative suggestions. //CC @thomaseizinger in case you are interested and have ideas.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -25,10 +25,7 @@ use std::borrow::Cow;
 /// # use prometheus_client::registry::Registry;
 /// #
 /// // Create a metric registry.
-/// //
-/// // Note the angle brackets to make sure to use the default (dynamic
-/// // dispatched boxed metric) for the generic type parameter.
-/// let mut registry = <Registry>::default();
+/// let mut registry = Registry::default();
 ///
 /// let counter: Counter = Counter::default();
 /// let gauge: Gauge = Gauge::default();
@@ -36,12 +33,12 @@ use std::borrow::Cow;
 /// registry.register(
 ///   "my_counter",
 ///   "This is my counter",
-///   Box::new(counter.clone()),
+///   counter.clone(),
 /// );
 /// registry.register(
 ///   "my_gauge",
 ///   "This is my gauge",
-///   Box::new(gauge.clone()),
+///   gauge.clone(),
 /// );
 ///
 /// # // Encode all metrics in the registry in the text format.
@@ -58,14 +55,14 @@ use std::borrow::Cow;
 /// # assert_eq!(expected, String::from_utf8(buffer).unwrap());
 /// ```
 #[derive(Debug)]
-pub struct Registry<M = Box<dyn crate::encoding::text::SendSyncEncodeMetric>> {
+pub struct Registry {
     prefix: Option<Prefix>,
     labels: Vec<(Cow<'static, str>, Cow<'static, str>)>,
-    metrics: Vec<(Descriptor, M)>,
-    sub_registries: Vec<Registry<M>>,
+    metrics: Vec<(Descriptor, Box<dyn Metric>)>,
+    sub_registries: Vec<Registry>,
 }
 
-impl<M> Default for Registry<M> {
+impl Default for Registry {
     fn default() -> Self {
         Self {
             prefix: None,
@@ -76,7 +73,7 @@ impl<M> Default for Registry<M> {
     }
 }
 
-impl<M> Registry<M> {
+impl Registry {
     /// Creates a new default [`Registry`] with the given prefix.
     pub fn with_prefix(prefix: impl Into<String>) -> Self {
         Self {
@@ -103,12 +100,17 @@ impl<M> Registry<M> {
     /// # use prometheus_client::metrics::counter::{Atomic as _, Counter};
     /// # use prometheus_client::registry::{Registry, Unit};
     /// #
-    /// let mut registry: Registry<Counter> = Registry::default();
-    /// let counter = Counter::default();
+    /// let mut registry = Registry::default();
+    /// let counter: Counter = Counter::default();
     ///
     /// registry.register("my_counter", "This is my counter", counter.clone());
     /// ```
-    pub fn register<N: Into<String>, H: Into<String>>(&mut self, name: N, help: H, metric: M) {
+    pub fn register<N: Into<String>, H: Into<String>>(
+        &mut self,
+        name: N,
+        help: H,
+        metric: impl Metric,
+    ) {
         self.priv_register(name, help, metric, None)
     }
 
@@ -124,8 +126,8 @@ impl<M> Registry<M> {
     /// # use prometheus_client::metrics::counter::{Atomic as _, Counter};
     /// # use prometheus_client::registry::{Registry, Unit};
     /// #
-    /// let mut registry: Registry<Counter> = Registry::default();
-    /// let counter = Counter::default();
+    /// let mut registry = Registry::default();
+    /// let counter: Counter = Counter::default();
     ///
     /// registry.register_with_unit(
     ///   "my_counter",
@@ -139,7 +141,7 @@ impl<M> Registry<M> {
         name: N,
         help: H,
         unit: Unit,
-        metric: M,
+        metric: impl Metric,
     ) {
         self.priv_register(name, help, metric, Some(unit))
     }
@@ -148,7 +150,7 @@ impl<M> Registry<M> {
         &mut self,
         name: N,
         help: H,
-        metric: M,
+        metric: impl Metric,
         unit: Option<Unit>,
     ) {
         let name = name.into();
@@ -164,7 +166,7 @@ impl<M> Registry<M> {
             labels: self.labels.clone(),
         };
 
-        self.metrics.push((descriptor, metric));
+        self.metrics.push((descriptor, Box::new(metric)));
     }
 
     // TODO: Update doc.
@@ -183,17 +185,17 @@ impl<M> Registry<M> {
     /// # use prometheus_client::metrics::counter::{Atomic as _, Counter};
     /// # use prometheus_client::registry::{Registry, Unit};
     /// #
-    /// let mut registry: Registry<Counter> = Registry::default();
+    /// let mut registry = Registry::default();
     ///
-    /// let subsystem_a_counter_1 = Counter::default();
-    /// let subsystem_a_counter_2 = Counter::default();
+    /// let subsystem_a_counter_1: Counter = Counter::default();
+    /// let subsystem_a_counter_2: Counter = Counter::default();
     ///
     /// let subsystem_a_registry = registry.sub_registry_with_prefix("subsystem_a");
     /// registry.register("counter_1", "", subsystem_a_counter_1.clone());
     /// registry.register("counter_2", "", subsystem_a_counter_2.clone());
     ///
-    /// let subsystem_b_counter_1 = Counter::default();
-    /// let subsystem_b_counter_2 = Counter::default();
+    /// let subsystem_b_counter_1: Counter = Counter::default();
+    /// let subsystem_b_counter_2: Counter = Counter::default();
     ///
     /// let subsystem_a_registry = registry.sub_registry_with_prefix("subsystem_b");
     /// registry.register("counter_1", "", subsystem_b_counter_1.clone());
@@ -239,7 +241,7 @@ impl<M> Registry<M> {
     }
 
     /// [`Iterator`] over all metrics registered with the [`Registry`].
-    pub fn iter(&self) -> RegistryIterator<M> {
+    pub fn iter(&self) -> RegistryIterator {
         let metrics = self.metrics.iter();
         let sub_registries = self.sub_registries.iter();
         RegistryIterator {
@@ -253,14 +255,14 @@ impl<M> Registry<M> {
 /// Iterator iterating both the metrics registered directly with the registry as
 /// well as all metrics registered with sub-registries.
 #[derive(Debug)]
-pub struct RegistryIterator<'a, M> {
-    metrics: std::slice::Iter<'a, (Descriptor, M)>,
-    sub_registries: std::slice::Iter<'a, Registry<M>>,
-    sub_registry: Option<Box<RegistryIterator<'a, M>>>,
+pub struct RegistryIterator<'a> {
+    metrics: std::slice::Iter<'a, (Descriptor, Box<dyn Metric>)>,
+    sub_registries: std::slice::Iter<'a, Registry>,
+    sub_registry: Option<Box<RegistryIterator<'a>>>,
 }
 
-impl<'a, M> Iterator for RegistryIterator<'a, M> {
-    type Item = &'a (Descriptor, M);
+impl<'a> Iterator for RegistryIterator<'a> {
+    type Item = &'a (Descriptor, Box<dyn Metric>);
 
     fn next(&mut self) -> Option<Self::Item> {
         if let Some(metric) = self.metrics.next() {
@@ -365,6 +367,45 @@ impl Unit {
     }
 }
 
+// TODO: Instead of depending on dynamic dispatch for all metrics, we could use
+// static dispatch for the most common ones and only fall back to dynamic
+// dispatch. That said, this needs to be proven performant in a benchmark first.
+/// Super trait representing an abstract Prometheus metric.
+#[cfg(not(feature = "protobuf"))]
+pub trait Metric:
+    crate::encoding::text::EncodeMetric + Send + Sync + std::fmt::Debug + 'static
+{
+}
+
+/// Super trait representing an abstract Prometheus metric.
+#[cfg(feature = "protobuf")]
+pub trait Metric:
+    crate::encoding::text::EncodeMetric
+    + crate::encoding::proto::EncodeMetric
+    + Send
+    + Sync
+    + std::fmt::Debug
+    + 'static
+{
+}
+
+#[cfg(not(feature = "protobuf"))]
+impl<T> Metric for T where
+    T: crate::encoding::text::EncodeMetric + Send + Sync + std::fmt::Debug + 'static
+{
+}
+
+#[cfg(feature = "protobuf")]
+impl<T> Metric for T where
+    T: crate::encoding::text::EncodeMetric
+        + crate::encoding::proto::EncodeMetric
+        + Send
+        + Sync
+        + std::fmt::Debug
+        + 'static
+{
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -372,8 +413,8 @@ mod tests {
 
     #[test]
     fn register_and_iterate() {
-        let mut registry: Registry<Counter> = Registry::default();
-        let counter = Counter::default();
+        let mut registry = Registry::default();
+        let counter: Counter = Counter::default();
         registry.register("my_counter", "My counter", counter);
 
         assert_eq!(1, registry.iter().count())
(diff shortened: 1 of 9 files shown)


adamchalmers:
Personally, I box all my metrics anyway, because I need to use many different types with one registry. So I'm okay with that tradeoff.

src/registry.rs
```diff
@@ -103,12 +100,17 @@ impl<M> Registry<M> {
...
+    /// let counter: Counter = Counter::default();
     ///
     /// registry.register("my_counter", "This is my counter", counter.clone());
     /// ```
-    pub fn register<N: Into<String>, H: Into<String>>(&mut self, name: N, help: H, metric: M) {
+    pub fn register<N: Into<String>, H: Into<String>>(
+        &mut self,
+        name: N,
+        help: H,
+        metric: impl Metric,
```
thomaseizinger, line 112: Would it make sense to take a reference here? I think the caller will always have to clone it, otherwise they are not gonna have a reference to it to actually use the metric, meaning we can do the cloning for them, therefore slightly simplifying the API.

src/registry.rs
```diff
@@ -183,17 +185,17 @@ impl<M> Registry<M> {
     /// # use prometheus_client::metrics::counter::{Atomic as _, Counter};
     /// # use prometheus_client::registry::{Registry, Unit};
     /// #
-    /// let mut registry: Registry<Counter> = Registry::default();
+    /// let mut registry = Registry::default();
     ///
-    /// let subsystem_a_counter_1 = Counter::default();
-    /// let subsystem_a_counter_2 = Counter::default();
+    /// let subsystem_a_counter_1: Counter = Counter::default();
+    /// let subsystem_a_counter_2: Counter = Counter::default();
```
thomaseizinger, line 191: These type ascriptions are unfortunate. Can we get rid of them somehow?

src/registry.rs
```diff
@@ -365,15 +367,54 @@ impl Unit {
...
+#[cfg(feature = "protobuf")]
+impl<T> Metric for T where
+    T: crate::encoding::text::EncodeMetric
+        + crate::encoding::proto::EncodeMetric
+        + Send
+        + Sync
+        + std::fmt::Debug
+        + 'static
+{
+}
```
thomaseizinger, line 407: This is a bit hacky but I guess it works? :D

thomaseizinger (commented):
Thanks for exploring this!

The problem domain we have here is basically the same as `serde`'s, thus I think the ideal solution might be a visitor pattern with an abstract data model.

That would make the actual `Encode` trait oblivious about the underlying format, same as `serde`'s `Serialize`. A registry could then pick a particular `Encoder` implementation and encode all metrics via double-dispatch.

This being said, `serde`'s design is obviously designed around having a plethora of actual serialization formats whereas here we only have two (I think?) so it might not worth it.

Thoughts?

mxinden:
> The problem domain we have here is basically the same as `serde`'s, thus I think the ideal solution might be a visitor pattern with an abstract data model.
> 
> That would make the actual `Encode` trait oblivious about the underlying format, same as `serde`'s `Serialize`. A registry could then pick a particular `Encoder` implementation and encode all metrics via double-dispatch.
> 
> This being said, `serde`'s design is obviously designed around having a plethora of actual serialization formats whereas here we only have two (I think?) so it might not worth it.
> 
> Thoughts?

The usage of serde is currently explored in https://github.com/prometheus/client_rust/issues/97. If I am not mistaken, this can happen in parallel, right? Serializing (serde) and storing (`Registry`) are orthogonal.

src/registry.rs
mxinden: My intention taking an owned value instead of a reference was to be explicit about the otherwise internal clone.

> I think the caller will always have to clone it, otherwise they are not gonna have a reference to it to actually use the metric

Not in the case of a constant metric, i.e. a metric that never changes. For such a metric the user does not need a handle to it, and can thus pass the only instance to `register`.

I would stick with the owned signature, though I don't feel strongly about it.

src/registry.rs
mxinden: Agreed. Though I am not sure how to do it differently.

src/registry.rs
mxinden: As far as I can tell, this is due to the default value of the `Counter` `A` trait parameter.

One thought I had was reducing the number of indirections through a `Counter::new` method, though unfortunately, the compiler can not infer `A` in such case either.

thomaseizinger:
> > The problem domain we have here is basically the same as `serde`'s, thus I think the ideal solution might be a visitor pattern with an abstract data model.
> > That would make the actual `Encode` trait oblivious about the underlying format, same as `serde`'s `Serialize`. A registry could then pick a particular `Encoder` implementation and encode all metrics via double-dispatch.
> > This being said, `serde`'s design is obviously designed around having a plethora of actual serialization formats whereas here we only have two (I think?) so it might not worth it.
> > Thoughts?
> 
> The usage of serde is currently explored in #97. If I am not mistaken, this can happen in parallel, right? Serializing (serde) and storing (`Registry`) are orthogonal.

You might have misunderstood me :)
I was pointing out the similarity between:

- Encoding formats in prometheus ("text" and "protobuf") and serde's serializers ("json", "toml", ..)
- The `Encode` trait and the `Serialize` trait

With this PR, we are facing the same problem as serde does: We want to have types (`Gauge`, `Counter`, etc) that should all know how to ~serialize~ encode themselves but we want them to be oblivious about the format (i.e. `text` vs `protobuf`).

We could solve this by copying serde's design of an abstract ~`Serializer`~ `Encoder` trait. `text::Encoder` is almost already the right abstraction, we just need to make it a trait and perhaps generalize it so it can do all the things we need to do for the protobuf format.

In other words, instead of having multiple `EncodeMetric` traits, we should have only one trait and multiple `Encoder` implementations that the user can choose between. The different metric types are essential our data model so the `Encoder` trait could look something like:

```rust
trait Encoder {
	fn encode_int_counter(&mut self, int_value: u64) -> Result<()>;
	fn encode_double_counter(&mut self, int_value: f64) -> Result<()>;
	fn encode_int_gauge(&mut self, int_value: i64) -> Result<()>;
	fn encode_labels(&mut self, labels: &[String]) -> Result<()>;
	// etc
}
```

Another design might be something like:

```rust
trait Encoder {
	fn start_encode_counter(&mut self) -> Result<CounterEncoder>;
	// etc
}

impl CounterEncoder<'encoder> {
	fn add_labels(&mut self, labels: &[String]);
	fn set_int_value(&mut self, int_value: u64);
	fn set_double_value(&mut self, double_value: f64);
	
	fn finish(self) -> Result<()>
}

impl Encode for Counter {
	fn encode<E>(&mut, encoder: &mut E) where E: Encoder -> Result<()> {
		let mut counter_encoder = encoder.start_encoder_counter();

		counter_encoder.add_labels(&["foo"]).set_int_value(self.value).finish()
	}
}
```

08d2:
I agree with @thomaseizinger that using `serde` directly is probably not the right approach.

Prometheus defines a set of types (counter, gauge, histogram, summary) which can be encoded. Core Prometheus defines one encoding format (text) and OpenMetrics another (binary) over those core types. No other valid encodings exist; there is no JSON encoding of a Prometheus counter, for example. The transformation between a Prometheus type and its encoded form is code defined by the format, and "custom" for the type. For example, a Rust type implementing a Prometheus gauge may have an OpenMetrics encoding which is different than its general Protobuf encoding.

mxinden:
:+1: for not adopting `serde`. Thanks for walking me through this. Appreciate the elaborate post above.

I like the idea of a single `EncodeMetric` trait which is then used by both `encoding::text` and `encoding::protobuf`.

I think this is orthogonal to this pull request. I.e. this pull request is about fixing `Registry` to use dynamic dispatching and thus dropping the trait parameter `M`.

I suggest we keep this pull request as is and in a follow-up pull request consolidate the two `EncodeMetric` traits into one.

Objections?

thomaseizinger:
> 👍 for not adopting `serde`. Thanks for walking me through this. Appreciate the elaborate post above.
> 
> I like the idea of a single `EncodeMetric` trait which is then used by both `encoding::text` and `encoding::protobuf`.
> 
> I think this is orthogonal to this pull request. I.e. this pull request is about fixing `Registry` to use dynamic dispatching and thus dropping the trait parameter `M`.
>
> I suggest we keep this pull request as is and in a follow-up pull request consolidate the two `EncodeMetric` traits into one.
> 
> Objections?

Fine with me.

I just thought a bit more about how we can adopt this design and we will have to take inspiration from https://github.com/dtolnay/miniserde as the static dispatch approach from `serde` does not work with trait objects as we want it where the registry doesn't know about the concrete metrics.

With saying that, could we get away with having an enum of all metrics and store that without boxing? :)

mxinden:
> With saying that, could we get away with having an enum of all metrics and store that without boxing? :)

Unfortunately not. Or I don't know how to do that. The reason being:

- `Counter`, `Gauge`, ... are generic over the underlying `AtomicXXX`, e.g. `Counter<AtomicU64>`.
- With `Family<M>`, where `M` can be any of `Counter`, `Gauge`, ... this results in a very high cardinality.

We could drop the generic parameters here as well, and only provide a limited set of `AtomicXXX`. Thus far I have not considered that worth the complexity.

thomaseizinger:
> > With saying that, could we get away with having an enum of all metrics and store that without boxing? :)
> 
> Unfortunately not. Or I don't know how to do that. The reason being:
> 
>     * `Counter`, `Gauge`, ... are generic over the underlying `AtomicXXX`, e.g. `Counter<AtomicU64>`.
> 
>     * `Family<M>`, where `M` can be any of `Counter`, `Gauge`, ...
> 
>     * A registry may store one or more of any permutation of the above, thus this results in a very high cardinality.

So a `Family` could just contain the enum again, right? It is a recursive structure like `serde_json::Value`.

Do we need to be generic over all atomics?

src/registry.rs
nox: Features are supposed to be additive, this is not additive.

thomaseizinger: Functionality-wise for the end user, it is additive. There is more discussion further down in how else we could solve this!

nox: If I care about only text encoding and my metric type implements only `crate::encoding::text::EncodeMetric`, and some dependency enables the `protobuf`, then my type won't implement `Metric` anymore.

thomaseizinger: It will. The one defined on line 382.

There will always be a `Metric` trait, just with different supertraits basd on the activated feature.

I can't think of a situation in which this breaks something but I agree that it is not clean.

thomaseizinger: Okay, yeah. It is technically breaking. But the set of prometheus metrics is fixed right? Why would you implement the trait yourself? What am I missing?

nox: I've already implemented [my own `Family`](https://github.com/nox/prometools/blob/11e2ce6248424b73643c0c9ec6f0e0c639abab54/src/serde/mod.rs#L19-L82) and [my own `Counter`](https://github.com/nox/prometools/blob/11e2ce6248424b73643c0c9ec6f0e0c639abab54/src/nonstandard.rs#L16-L19).

mxinden:
> So a `Family` could just contain the enum again, right? It is a recursive structure like `serde_json::Value`.

Correct.

> Do we need to be generic over all atomics?

Maybe. I am undecided on it.

Also `Family` is not only generic over the metric `M`, but also over the label set `S` and the constructor `C`. We would need to replace these with trait objects. I am not sure that would be a good idea.

src/registry.rs
mxinden: Thanks for raising this @nox.

I think the way to go is a single `EncodeMetric` trait which then allows `encoding:text` and `encoding::protobuf` to traverse a metric via the visitor patter as suggested below by @thomaseizinger.

thomaseizinger:
> > So a `Family` could just contain the enum again, right? It is a recursive structure like `serde_json::Value`.
> 
> Correct.
> 
> > Do we need to be generic over all atomics?
> 
> Maybe. I am undecided on it.
> 
> Also `Family` is not only generic over the metric `M`, but also over the label set `S` and the constructor `C`. We would need to replace these with trait objects. I am not sure that would be a good idea.

Fair enough. Yeah it would just be another optimization. We should be fine if we box the entire metric up.

mxinden:
> ```rust
> impl Encode for Counter {
> 	fn encode<E>(&mut, encoder: &mut E) where E: Encoder -> Result<()> {
> 		let mut counter_encoder = encoder.start_encoder_counter();
> 
> 		counter_encoder.add_labels(&["foo"]).set_int_value(self.value).finish()
> 	}
> }
> ```

Following up on above's suggestion in https://github.com/prometheus/client_rust/pull/100#issuecomment-1266341578.

The `Encode` trait as described above is not trait object safe. Thus we could not store a `Box<dyn Encode>` in a `Registry`.

mxinden:
While I do see value in users being able to bring their own metrics, I don't think users will bring their own output format. Thus we can likely get around the above trait object safety issue by multiplexing output formats via an `enum`. I am working on a proof-of-concept. More to come.

thomaseizinger:
> While I do see value in users being able to bring their own metrics, I don't think users will bring their own output format. Thus we can likely get around the above trait object safety issue by multiplexing output formats via an `enum`. I am working on a proof-of-concept. More to come.

Instead of an enum, `&mut dyn Encoder` might work too? The issue I see with an enum is that it might again be hard to feature-flag the protobuf implementation. We would have to wrap the enum and make `Encoder` an opaque struct to make sure users don't depend on the variants.

mxinden:
> Instead of an enum, `&mut dyn Encoder` might work too?

In that case none of the methods on `Encoder` are allowed to have generic type parameters. Though we want the `Encoder` to be able to encode some generic type that implements `EncodeLabelSet`. See error below.

I have not yet found a way around this issue.

```
error[E0038]: the trait `encoding::MetricEncoder` cannot be made into an object
  --> src/encoding.rs:29:36
   |
29 |     fn encode(&self, encoder: &mut dyn MetricEncoder) -> Result<(), std::fmt::Error> {
   |                                    ^^^^^^^^^^^^^^^^^ `encoding::MetricEncoder` cannot be made into an object
   |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
  --> src/encoding.rs:45:8
   |
39 | pub trait MetricEncoder<'a> {
   |           ------------- this trait cannot be made into an object...
...
45 |     fn with_label_set<'b, S: EncodeLabelSet>(
   |        ^^^^^^^^^^^^^^ ...because method `with_label_set` has generic type parameters
   = help: consider moving `with_label_set` to another trait
```

---

> The issue I see with an enum is that it might again be hard to feature-flag the protobuf implementation. We would have to wrap the enum and make `Encoder` an opaque struct to make sure users don't depend on the variants.

True. Though that might be worth it.

mxinden:
https://github.com/prometheus/client_rust/pull/105/ implements the above suggested `serde` style encoding using two traits, `EncodeMetric` (see serde's `Serialize`) and `MetricEncoder` (see serde's `Serializer`).

This addresses the issue raised by @nox, namely the conflict with Rust's additive features.

Feedback is more than welcome.

08d2:
> While I do see value in users being able to bring their own metrics

I'm probably missing something here, but this doesn't (immediately) make sense to me. Prometheus metrics are types, not traits. A user type wouldn't _implement_ a counter, it would _update_ a counter: has-a, rather than is-a. Right?

thomaseizinger:
> > While I do see value in users being able to bring their own metrics
> 
> I'm probably missing something here, but this doesn't (immediately) make sense to me. Prometheus metrics are types, not traits. A user type wouldn't _implement_ a counter, it would _update_ a counter: has-a, rather than is-a. Right?

I don't quite understand what you are getting at. Every metric needs to implement `EncodeMetric` in the suggested solution. How is that related to has-a vs is-a?

nox:
> I'm probably missing something here, but this doesn't (immediately) make sense to me. Prometheus metrics are types, not traits. A user type wouldn't _implement_ a counter, it would _update_ a counter: has-a, rather than is-a. Right?

As mentioned in [this previous comment](https://github.com/prometheus/client_rust/pull/100#discussion_r992174609), I've already implemented my own counter and my own family. That they wrap types from this crate is an implementation detail, the important part is that my custom types implement `EncodeMetric` without delegating to the prometheus-client types they wrap.

08d2:
@thomaseizinger 

> > I'm probably missing something here, but this doesn't (immediately) make sense to me. Prometheus metrics are types, not traits. A user type wouldn't _implement_ a counter, it would _update_ a counter: has-a, rather than is-a. Right?
> 
> I don't quite understand what you are getting at. Every metric needs to implement `EncodeMetric` in the suggested solution. How is that related to has-a vs is-a?

Ah, apologies for being vague. Let me try again: if you're writing a Prometheus client library (i.e. this crate) then the idea is that you're providing "every metric" yourself, each as a specific concrete type. Metrics are precisely specified, and their behavior isn't (or, shouldn't) be up to interpretation by implementations. 

@nox 

> As mentioned in https://github.com/prometheus/client_rust/pull/100#discussion_r992174609, I've already implemented my own counter and my own family. That they wrap types from this crate is an implementation detail, the important part is that my custom types implement EncodeMetric without delegating to the prometheus-client types they wrap.

Your counter implementation violates the Prometheus spec, I think, right? So it isn't really a counter, even though it may satisfy the relevant interface in this client library. This is an example of the kind of outcome that I claim should be specifically prevented.

YMMV 🤷

nox:
> Your counter implementation violates the Prometheus spec, I think, right?

The OpenMetrics spec, sure. Other *prometheus* clients [so far](https://github.com/prometheus/client_golang/issues/725) [don't care](https://github.com/prometheus/client_ruby/issues/189) about suffixes.

If the ability to make my own unsuffixed counter goes away, I'll simply not use this crate anymore.

thomaseizinger:
> @thomaseizinger 
> 
> > > I'm probably missing something here, but this doesn't (immediately) make sense to me. Prometheus metrics are types, not traits. A user type wouldn't _implement_ a counter, it would _update_ a counter: has-a, rather than is-a. Right?
> > 
> > I don't quite understand what you are getting at. Every metric needs to implement `EncodeMetric` in the suggested solution. How is that related to has-a vs is-a?
> 
> Ah, apologies for being vague. Let me try again: if you're writing a Prometheus client library (i.e. this crate) then the idea is that you're providing "every metric" yourself, each as a specific concrete type. Metrics are precisely specified, and their behavior isn't (or, shouldn't) be up to interpretation by implementations. Users shouldn't be bringing their own types. They should be using yours.

Yes, I agree with you. The way I see it is that "bring your own metric" is likely to be possible but I don't think it needs to be a goal of the library.

08d2:
> > Your counter implementation violates the Prometheus spec, I think, right?
> 
> The OpenMetrics spec, sure. Other _prometheus_ clients [so far](https://github.com/prometheus/client_golang/issues/725) [don't care](https://github.com/prometheus/client_ruby/issues/189) about suffixes.
> 
> If the ability to make my own unsuffixed counter goes away, I'll simply not use this crate anymore.

This is at least superficially confusing. Why? A counter in the context of Prometheus is, by definition, suffixed.

nox:
> > > Your counter implementation violates the Prometheus spec, I think, right?
> > 
> > 
> > The OpenMetrics spec, sure. Other _prometheus_ clients [so far](https://github.com/prometheus/client_golang/issues/725) [don't care](https://github.com/prometheus/client_ruby/issues/189) about suffixes.
> 
> The docs on naming state that
> 
> > an accumulating count has total as a suffix, in addition to the unit if applicable
> 
> https://prometheus.io/docs/practices/naming/#metric-names

Literally not a spec, this document starts with:

> The metric and label conventions presented in this document are not required for using Prometheus, but can serve as both a style-guide and a collection of best practices. *Individual organizations may want to approach some of these practices, e.g. naming conventions, differently.*

(Emphasis mine)

08d2:
> Literally not a spec, this document starts with:
> 
> > The metric and label conventions presented in this document are not required for using Prometheus, but can serve as both a style-guide and a collection of best practices. _Individual organizations may want to approach some of these practices, e.g. naming conventions, differently._
> 
> (Emphasis mine)

As this client library exists in the Prometheus org, I assume it will abide and enforce Prometheus best practices. If that's not the case, then I suppose it should live somewhere else.

nox:
> But whether spec or suggestion, MUST or SHOULD, the things under discussion here, suffixing certainly included, represent no meaningful hardship to users, and carry no cost versus alternatives. Absent specific cause, there's no reason not to play by the rules.

I've already explained that we have dozens of dozens of dashboards using unsuffixed metrics and forking this crate will be cheaper than migrating those dashboards.

mxinden:
I am closing here. The goal of this pull request has been achieved through #105 and released as an alpha via https://github.com/prometheus/client_rust/pull/116.

Thanks for the input everyone. Would appreciate folks testing the new alpha and leaving feedback on https://github.com/prometheus/client_rust/pull/116 or as separate GitHub issues.

nox:
> I've already explained that we have dozens of dozens of dashboards using unsuffixed metrics and forking this crate will be cheaper than migrating those dashboards.

You completely removed the ability to define custom metrics to avoid the `_total` suffix on counters, that's really unfortunate.

mxinden:
I am sorry for the trouble @nox. I decided to follow the OpenMetrics specification. In the long run I hope for the ecosystem (e.g. your dashboards) to converge on that specification. In the meantime, I don't see a way around forking.

nox:
> I am sorry for the trouble @nox. I decided to follow the OpenMetrics specification. In the long run I hope for the ecosystem (e.g. your dashboards) to converge on that specification. In the meantime, I don't see a way around forking.

Thanks, we will stay on prometheus-client 0.18 for now.

PR #102, opened 2022-10-13
Fix Family::get_or_create

nox:
The method should not overwrite an existing key after it obtains the write lock.

Consider the following scenario with `Family<LabelSet, Gauge>` used by threads A and B:

1. A and B both call `family.get_or_insert(&label_set)`
2. A and B both acquire read lock and finds no key
3. A gets write lock and inserts a new gauge with value 0
4. A increments returned gauge to 1
5. B gets write lock and inserts a new gauge with value 0
6. B increments returned gauge to 1
7. A decrements gauge back to 0
8. B decrements gauge which now overflows to `u64::MAX`

### DIFF
--- a/src/metrics/family.rs
+++ b/src/metrics/family.rs
@@ -3,7 +3,7 @@
 //! See [`Family`] for details.
 
 use super::{MetricType, TypedMetric};
-use parking_lot::{MappedRwLockReadGuard, RwLock, RwLockReadGuard};
+use parking_lot::{MappedRwLockReadGuard, RwLock, RwLockReadGuard, RwLockWriteGuard};
 use std::collections::HashMap;
 use std::sync::Arc;
 
@@ -223,11 +223,14 @@ impl<S: Clone + std::hash::Hash + Eq, M, C: MetricConstructor<M>> Family<S, M, C
         }
 
         let mut write_guard = self.metrics.write();
-        write_guard.insert(label_set.clone(), self.constructor.new_metric());
 
-        drop(write_guard);
+        write_guard
+            .entry(label_set.clone())
+            .or_insert_with(|| self.constructor.new_metric());
 
-        RwLockReadGuard::map(self.metrics.read(), |metrics| {
+        let read_guard = RwLockWriteGuard::downgrade(write_guard);
+
+        RwLockReadGuard::map(read_guard, |metrics| {
             metrics
                 .get(label_set)
                 .expect("Metric to exist after creating it.")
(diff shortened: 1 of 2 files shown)


mxinden (commented):
Oh silly me.

Great catch. Thanks for the patch.

Can you include a changelog entry?

Out of curiosity, did you hit this race condition, or found it while reading?

nox:
> Out of curiosity, did you hit this race condition, or found it while reading?

We discovered this the hard way with gauges reporting increments of 18.4 quintillions after making the switch in 3 different repos.

nox:
I'll add a changelog entry. Can you release 0.18.1 with that fix?

src/metrics/family.rs
```diff
@@ -222,10 +222,10 @@ impl<S: Clone + std::hash::Hash + Eq, M, C: MetricConstructor<M>> Family<S, M, C
...
-        let mut write_guard = self.metrics.write();
-        write_guard.insert(label_set.clone(), self.constructor.new_metric());
-
-        drop(write_guard);
+        self.metrics
+            .write()
+            .entry(label_set.clone())
+            .or_insert_with(|| self.constructor.new_metric());
 
         RwLockReadGuard::map(self.metrics.read(), |metrics| {
```
mxinden, line 230: With `Family::remove` there is actually a second race condition.

In the scenarios where we have two threads:

A: Call `Family::get_or_create` with a new label set.
    1. Acquiring the read lock. Nothing there.
    2. Acquiring the write lock. Inserting the new metric.
    3. Being preempted by the OS.
B: Call `Family::remove` with the above label set.
A: Acquiring the read lock for the second (last) time, `.expect()`ing the previously inserted metric to be there.

This can be prevented by leveraging `RwLockWriteGuard::downgrade` of the previously acquired write lock instead of doing acquiring another read lock via `self.metrics.read`.

Does the above make sense?

mxinden (commented):
> > Out of curiosity, did you hit this race condition, or found it while reading?
> 
> We discovered it the hard way with gauges reporting increments of 18.4 quintillions after making the switch in 3 repos.

Sorry to hear that. My bad. Thanks for debugging.

Do you have some time to tackle the second race condition as well? If not, I will give it a shot tomorrow.

> I'll add a changelog entry. Can you release 0.18.1 with that fix?

Yes, I will do that as soon as we have both race conditions fixed. See below.

src/metrics/family.rs
nox: It makes sense yes. The patch now uses `RwLockWriteGuard::downgrade`.

mxinden:
Released with https://github.com/prometheus/client_rust/pull/103 as `v0.18.1`. Merged into `master` via https://github.com/prometheus/client_rust/pull/104.

I am closing here.

Again, thanks @nox for debugging and thanks for providing a fix. Great work.

PR #105, opened 2022-10-16
encoding/: Adopt serde style encoding

mxinden:
Adopt encoding style similar to serde. `EncodeXXX` for a type that can be
encoded (see serde's `Serialize`) and `XXXEncoder` for a supported encoding i.e.
text and protobuf (see serde's `Serializer`).

- Compatible with Rust's additive features. Enabling the `protobuf` feature does
not change type signatures.
- `EncodeMetric` is trait object safe, and thus `Registry` can use dynamic dispatch.
- Implement a single encoding trait `EncodeMetric` per metric type instead of
one implementation per metric AND per encoding format.
- Leverage `std::fmt::Write` instead of `std::io::Write`. The OpenMetrics text
format has to be unicode compliant. The OpenMetrics Protobuf format requires
labels to be valid strings.

Builds on top of https://github.com/prometheus/client_rust/pull/100.

Status: It compiles. Still a couple of rough edges. Tests don't yet compile.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -2,6 +2,559 @@
 
 pub use prometheus_client_derive_encode::*;
 
+use crate::metrics::exemplar::Exemplar;
+use crate::metrics::MetricType;
+use std::borrow::Cow;
+use std::collections::HashMap;
+use std::fmt::Write;
+use std::ops::Deref;
+
 #[cfg(feature = "protobuf")]
-pub mod proto;
+pub mod protobuf;
 pub mod text;
+
+/// Trait implemented by each metric type, e.g.
+/// [`Counter`](crate::metrics::counter::Counter), to implement its encoding in
+/// the OpenMetric text format.
+pub trait EncodeMetric {
+    /// Encode the given instance in the OpenMetrics text encoding.
+    fn encode(&self, encoder: MetricEncoder<'_, '_>) -> Result<(), std::fmt::Error>;
+
+    /// The OpenMetrics metric type of the instance.
+    // One can not use [`TypedMetric`] directly, as associated constants are not
+    // object safe and thus can not be used with dynamic dispatching.
+    fn metric_type(&self) -> MetricType;
+}
+
+impl EncodeMetric for Box<dyn EncodeMetric> {
+    fn encode(&self, encoder: MetricEncoder) -> Result<(), std::fmt::Error> {
+        self.deref().encode(encoder)
+    }
+
+    fn metric_type(&self) -> MetricType {
+        self.deref().metric_type()
+    }
+}
+
+/// Encoder for a metric.
+///
+// `MetricEncoder` does not take a trait parameter for `writer` and `labels`
+// because `EncodeMetric` which uses `MetricEncoder` needs to be usable as a
+// trait object in order to be able to register different metric types with a
+// `Registry`. Trait objects can not use type parameters.
+//
+// TODO: Alternative solutions to the above are very much appreciated.
+#[derive(Debug)]
+pub struct MetricEncoder<'a, 'b>(MetricEncoderInner<'a, 'b>);
+
+#[derive(Debug)]
+enum MetricEncoderInner<'a, 'b> {
+    Text(text::MetricEncoder<'a, 'b>),
+
+    #[cfg(feature = "protobuf")]
+    Protobuf(protobuf::MetricEncoder<'a>),
+}
+
+impl<'a, 'b> From<text::MetricEncoder<'a, 'b>> for MetricEncoder<'a, 'b> {
+    fn from(e: text::MetricEncoder<'a, 'b>) -> Self {
+        Self(MetricEncoderInner::Text(e))
+    }
+}
+
+#[cfg(feature = "protobuf")]
+impl<'a, 'b> From<protobuf::MetricEncoder<'a>> for MetricEncoder<'a, 'b> {
+    fn from(e: protobuf::MetricEncoder<'a>) -> Self {
+        Self(MetricEncoderInner::Protobuf(e))
+    }
+}
+
+macro_rules! for_both_mut {
+    ($self:expr, $inner:ident, $pattern:pat, $fn:expr) => {
+        match &mut $self.0 {
+            $inner::Text($pattern) => $fn,
+            #[cfg(feature = "protobuf")]
+            $inner::Protobuf($pattern) => $fn,
+        }
+    };
+}
+
+macro_rules! for_both {
+    ($self:expr, $inner:ident, $pattern:pat, $fn:expr) => {
+        match $self.0 {
+            $inner::Text($pattern) => $fn,
+            #[cfg(feature = "protobuf")]
+            $inner::Protobuf($pattern) => $fn,
+        }
+    };
+}
+
+impl<'a, 'b> MetricEncoder<'a, 'b> {
+    /// Encode a counter.
+    pub fn encode_counter<
+        S: EncodeLabelSet,
+        CounterValue: EncodeCounterValue,
+        ExemplarValue: EncodeExemplarValue,
+    >(
+        &mut self,
+        v: CounterValue,
+        exemplar: Option<&Exemplar<S, ExemplarValue>>,
+    ) -> Result<(), std::fmt::Error> {
+        for_both_mut!(self, MetricEncoderInner, e, e.encode_counter(v, exemplar))
+    }
+
+    /// Encode a gauge.
+    pub fn encode_gauge(&mut self, v: impl EncodeGaugeValue) -> Result<(), std::fmt::Error> {
+        for_both_mut!(self, MetricEncoderInner, e, e.encode_gauge(v))
+    }
+
+    /// Encode an info.
+    pub fn encode_info(&mut self, label_set: &impl EncodeLabelSet) -> Result<(), std::fmt::Error> {
+        for_both_mut!(self, MetricEncoderInner, e, e.encode_info(label_set))
+    }
+
+    /// Encode a histogram.
+    pub fn encode_histogram<S: EncodeLabelSet>(
+        &mut self,
+        sum: f64,
+        count: u64,
+        buckets: &[(f64, u64)],
+        exemplars: Option<&HashMap<usize, Exemplar<S, f64>>>,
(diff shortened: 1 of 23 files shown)


src/encoding.rs
```diff
@@ -2,6 +2,502 @@
...
+use crate::metrics::MetricType;
+use std::borrow::Cow;
+use std::fmt::Write;
+use std::ops::Deref;
+
 #[cfg(feature = "protobuf")]
 pub mod proto;
 pub mod text;
+
+// TODO: Rename to MetricEncode to be consistent with MetricEncoder.
```
thomaseizinger, line 16: Everything we do are metrics in here, yes? Can we drop the `Metric` prefix?

src/encoding.rs
```diff
@@ -2,6 +2,502 @@
...
+pub trait EncodeLabelKey {
+    fn encode(&self, encoder: &mut LabelKeyEncoder) -> Result<(), std::fmt::Error>;
+}
+
+pub enum LabelKeyEncoder<'a> {
+    Text(text::LabelKeyEncoder<'a>),
+    Protobuf(proto::LabelKeyEncoder<'a>),
+}
+
+impl<'a> From<text::LabelKeyEncoder<'a>> for LabelKeyEncoder<'a> {
```
thomaseizinger, line 298: I'd expect that these are only used within this crate (and not exported publicly) so I am not sure we need these impls.

src/encoding.rs
```diff
@@ -2,6 +2,502 @@
...
+        Ok(())
+    }
+}
+
+impl<K: EncodeLabelKey, V: EncodeLabelValue> EncodeLabel for (K, V) {
+    fn encode(&self, mut encoder: LabelEncoder) -> Result<(), std::fmt::Error> {
+        let (key, value) = self;
+
+        let mut label_key_encoder = encoder.encode_label_key();
+        key.encode(&mut label_key_encoder)?;
```
thomaseizinger, line 375: In serde, I believe this is flipped: The serializer has functions for writing certain types, like `Serializer::serialize_str`: https://docs.rs/serde/latest/serde/trait.Serializer.html#tymethod.serialize_str

Would it make sense to do the same thing here? Our "data format" are metrics so I think it would make sense if an encoder has a function "write_label" and you can pass a string to it.

I still need to dig into this in detail so excuse if it doesn't make any sense!

src/metrics/counter.rs
```diff
@@ -171,6 +175,62 @@ impl<N, A> TypedMetric for Counter<N, A> {
...
+}
+
+fn encode_counter_with_maybe_exemplar<S, CounterValue, ExemplarValue>(
+    value: CounterValue,
+    exemplar: Option<&crate::metrics::exemplar::Exemplar<S, ExemplarValue>>,
+    mut encoder: MetricEncoder,
+) -> Result<(), std::fmt::Error>
+where
+    S: EncodeLabelSet,
+    CounterValue: EncodeValue,
```
thomaseizinger, line 218: Instead of dispatching to the value here, we could have one function on `MetricEncoder` per type and we'd probably just need to `macro_rules` the counter definition for each supported value type.

Not sure if worth it but I think it would reduce some indirection in dispatching between all these things.

thomaseizinger (commented):
Did an initial, high-level pass!

This looks like a good step forward, I still need to dig into the details :)
Thanks for giving this a shot!

thomaseizinger:
I had a closer look at this and I found the the decomposition of the functions on `MetricEncoder` a bit confusing to understand. What do you think about an approach where we provide more concrete functions on `MetricEncoder` like https://github.com/prometheus/client_rust/commit/120cfb0e97e3f7853a4ac8cdd1f8e4418e533f65?

I think this could drastically simplify, how often we need to dispatch and e.g. also solves the issue that the protobuf encoding does not know about suffixes. On the flip side, this will decrease the flexibility a bit because users cannot make arbitrary suffixes anymore. This feature is a bit of an illusion anyway because it only works with the text encoding.

thomaseizinger:
> > I had a closer look at this and I found the the decomposition of the functions on `MetricEncoder` a bit confusing to understand. What do you think about an approach where we provide more concrete functions on `MetricEncoder` like [120cfb0](https://github.com/prometheus/client_rust/commit/120cfb0e97e3f7853a4ac8cdd1f8e4418e533f65)?
> 
> Thanks for looking into this!
> 
> I think that makes sense. Would you agree with me suggesting that we can do this as a follow up @thomaseizinger? I would like this pull request to make the architectural change only. Breaking individual methods can happen in a one-by-one case later on, ideally in a deprecate-then-remove style in case users already depend on it.

I suggested it because I think it would really simplify the implementation and thus ease the review of this PR.

I get the desire to keep the PR small. Unfortunately, we have to add a lot of code to retain some of the interfaces. I think this PR could be smaller if we just break those.

mxinden:
> I had a closer look at this and I found the the decomposition of the functions on `MetricEncoder` a bit confusing to understand. What do you think about an approach where we provide more concrete functions on `MetricEncoder` like [120cfb0](https://github.com/prometheus/client_rust/commit/120cfb0e97e3f7853a4ac8cdd1f8e4418e533f65)?

Gave this more thought and tested it out. See https://github.com/prometheus/client_rust/pull/105/commits/e030f76aad4e8b29b1398eac77a9fa3fbc954ff1. As you predicted, this simplifies things a lot.

As always, thanks @thomaseizinger for the great suggestion.

There are still a couple of TODOs, but I see light at the end of the tunnel.

thomaseizinger:
> > I had a closer look at this and I found the the decomposition of the functions on `MetricEncoder` a bit confusing to understand. What do you think about an approach where we provide more concrete functions on `MetricEncoder` like [120cfb0](https://github.com/prometheus/client_rust/commit/120cfb0e97e3f7853a4ac8cdd1f8e4418e533f65)?
> 
> Gave this more thought and tested it out. See [e030f76](https://github.com/prometheus/client_rust/commit/e030f76aad4e8b29b1398eac77a9fa3fbc954ff1). As you predicted, this simplifies things a lot.

That is great to hear! I think we can do even better though. Let me know what you think of https://github.com/thomaseizinger/client_rust/commit/d2ad9df2deb99191f3b67d2022e7f75c05b9c48d.

mxinden:
> > > I had a closer look at this and I found the the decomposition of the functions on `MetricEncoder` a bit confusing to understand. What do you think about an approach where we provide more concrete functions on `MetricEncoder` like [120cfb0](https://github.com/prometheus/client_rust/commit/120cfb0e97e3f7853a4ac8cdd1f8e4418e533f65)?
> > 
> > 
> > Gave this more thought and tested it out. See [e030f76](https://github.com/prometheus/client_rust/commit/e030f76aad4e8b29b1398eac77a9fa3fbc954ff1). As you predicted, this simplifies things a lot.
> 
> That is great to hear! I think we can do even better though. Let me know what you think of [thomaseizinger@d2ad9df](https://github.com/thomaseizinger/client_rust/commit/d2ad9df2deb99191f3b67d2022e7f75c05b9c48d).

Wonderful simplification. Thanks for proposing a patch. Only issue I see is that it makes certain invalid states possible at compile time. See https://github.com/thomaseizinger/client_rust/commit/d2ad9df2deb99191f3b67d2022e7f75c05b9c48d#r90329584.

thomaseizinger:
I sent a PR here: https://github.com/mxinden/client_rust/pull/1

howardjohn:
Is there a reason this is adopting a serde *style* rather than directly using Serde (https://github.com/prometheus/client_rust/issues/97)? The motivation behind that would be that serde already has a large ecosystem behind it, with plenty of helpers to construct Serialize implementations like `#[flatten]`, etc

thomaseizinger:
Not every type that implements `Serialize` is a valid metric.

Maybe we could use serde internally but even that is tricky because protobuf doesn't map cleanly to serde so I am struggling to see the benefit!

How do you expect the ecosystem support of serde to benefit this prometheus client implementation?

howardjohn:
That makes sense, thanks for the explanation. My main interest was from more advanced `derive` support; in particular I want `#[serde(flatten)]` for now, but possibly other customizations in the future. While these could be solved by alternative mechanisms, it seemed nice to just piggy-back on Serde

08d2:
Each Prometheus metric encoding would map to a unique Serde data format, all distinct from e.g. raw Protobufs or whatever. Only a well-defined set of encodings (data formats) would be acceptable representations of metrics. One couldn't `derive` an arbitrary data format for a type and expect it to work. (If this was already understood, apologies!)

thomaseizinger:
> That makes sense, thanks for the explanation. My main interest was from more advanced `derive` support; in particular I want `#[serde(flatten)]` for now, but possibly other customizations in the future. While these could be solved by alternative mechanisms, it seemed nice to just piggy-back on Serde

Can you please explain further what you are trying to achieve? Like, post the type you are trying to write (with serde annotations) and the corresponding prometheus text encoding that you are expecting.

thomaseizinger:
Okay, thanks for sharing this. It makes sense that we support something like this but we can't do it via `serde`, simply because there is no way of how we can possible map all types that are `Serialize` to a valid prometheus text encoding.

With the features introduced in this PR, it will be possible to implement a `#[labels(flatten)]` attribute in the custom derive. Until then, you can workaround it by making a constructor that accepts `Common` and flattens out the fields manually. That is only a few more LoC and provides roughly the same convenience upon constructing the label set.

howardjohn:
In real world we have about 10 metrics with 15 common labels so it ends up being a bit of duplication. Anyhow, I don't want to derail this PR. Thanks for the help and explaination, I'll open an issue to track adding a flatten attribute.

mxinden:
> With the features introduced in this PR, it will be possible to implement a `#[labels(flatten)]` attribute in the custom derive.

That would be my preference as well.

@howardjohn let's discuss how to proceed once you created the issue.

mxinden:
I will merge here. I tested this in conjunction with #82 on https://github.com/mxinden/kademlia-exporter/. While #82 still needs some work, this is ready to go. I suggest we cut an `alpha` in between to allow folks to test this patch and the new protobuf encoding from @ackintosh.

howardjohn:
I opened https://github.com/prometheus/client_rust/issues/117 to discuss further. Thanks everyone. I may be able to submit a PR myself if I can find time + figure it out

Issue #117, opened 2022-12-12
Add `#[flatten]` directive for derivation of nested structs


howardjohn:
Given the structs like:

```rust
struct Common {
  field_a: String,
  field_b: String,
  field_c: String
}

#[derive(Encode)]
struct Metric1 {
  unique_field: String
  #[flatten]
  common: Common
}

#[derive(Encode)]
struct Metric2 {
  unique_field_2: String
  #[serde(flatten)]
  common: Common
}
```

I want the resulting metrics:

```
metric1{field_a="foo", field_b="bar", field_c="baz",unique_field="something"}
metric2{field_a="foo", field_b="bar", field_c="baz",unique_field_2="something_else"}
```

(Note: in real world Common and the number of metrics are  much larger)

This would match the semantics of `serde(flatten)`

See https://github.com/prometheus/client_rust/pull/105#issuecomment-1340337044 for more discussion

mxinden:
I have had that same use-case in rust-libp2p before. I am in favor of adding this to `prometheus-client-derive-encode`.

I foresee having the capacity to implement this any time soon. Focusing all my capacity on #74. @howardjohn are you interested in making the contribution?

howardjohn:
@mxinden I sent a PR in https://github.com/prometheus/client_rust/pull/118. it might need some improvements but it works in the basic case right now, at least.

PR #118, opened 2022-12-12
feat(derive): Introduce flatten attribute

howardjohn:
Fixes https://github.com/prometheus/client_rust/issues/117

This introduces a new attribute to the derive code for to flatten. This matches https://serde.rs/attr-flatten.html.

This seems to work, but I have some concern that if there are duplicate fields we end up encoding all of them which may make things go awry. Checking this is hard since we would need to lookup the type, which I am not sure how to do...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Introduce `Collector` abstraction allowing users to provide additional metrics
   and their description on each scrape. See [PR 82].
 - Introduce a `#[prometheus(flatten)]` attribute which can be used when deriving `EncodeLabelSet`, allowing
-  a nested struct the be flattened during encoding. See [PR 118].
+  a nested struct to be flattened during encoding. See [PR 118].
+  For example:
+  ```rust
+  #[derive(EncodeLabelSet, Hash, Clone, Eq, PartialEq, Debug)]
+  struct CommonLabels {
+      a: u64,
+      b: u64,
+  }
+  #[derive(EncodeLabelSet, Hash, Clone, Eq, PartialEq, Debug)]
+  struct Labels {
+      unique: u64,
+      #[prometheus(flatten)]
+      common: CommonLabels,
+  }
+
+  // Would result in `my_metric{a="42",b="42",unique="42"} 42
+  ```
 
 [PR 82]: https://github.com/prometheus/client_rust/pull/82
 [PR 118]: https://github.com/prometheus/client_rust/pull/118
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Introduce `Collector` abstraction allowing users to provide additional metrics
   and their description on each scrape. See [PR 82].
+- Introduce a `#[prometheus(flatten)]` attribute which can be used when deriving `EncodeLabelSet`, allowing
+  a nested struct the be flattened during encoding. See [PR 118].
 
 [PR 82]: https://github.com/prometheus/client_rust/pull/82
+[PR 118]: https://github.com/prometheus/client_rust/pull/118
 
 ## [0.19.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Introduce `Collector` abstraction allowing users to provide additional metrics
   and their description on each scrape. See [PR 82].
+
 - Introduce a `#[prometheus(flatten)]` attribute which can be used when deriving `EncodeLabelSet`, allowing
   a nested struct to be flattened during encoding. See [PR 118].
+
   For example:
+
   ```rust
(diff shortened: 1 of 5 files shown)


mxinden:
> This seems to work, but I have some concern that if there are duplicate fields we end up encoding all of them which may make things go awry. Checking this is hard since we would need to lookup the type, which I am not sure how to do...

If I am not mistaken, `serde` isn't able to handle duplicate fields at compile time either.

derive-encode/src/lib.rs
```diff
@@ -22,22 +22,35 @@ pub fn derive_encode_label_set(input: TokenStream) -> TokenStream {
...
-                        let mut label_encoder = encoder.encode_label();
-                        let mut label_key_encoder = label_encoder.encode_label_key()?;
-                        EncodeLabelKey::encode(&#ident_string, &mut label_key_encoder)?;
-
-                        let mut label_value_encoder = label_key_encoder.encode_label_value()?;
-                        EncodeLabelValue::encode(&self.#ident, &mut label_value_encoder)?;
-
-                        label_value_encoder.finish()?;
+                        .find(|a| a.path.is_ident("prometheus"))
+                        .map(|a| a.parse_args::<syn::Ident>().unwrap().to_string() == "flatten")
```
mxinden, line 29: Today the derive macro does not support any attributes. With that in mind, what do you think of providing an error message here in case a user provides anything other than `flatten`? E.g. in the case of `flaaaaatten` it would be helpful for the user to get an error message along the lines of: "Provided attribute `flaaaaatten` but only `flatten` is supported.". What do you think?

derive-encode/src/lib.rs
```diff
@@ -22,22 +22,35 @@ pub fn derive_encode_label_set(input: TokenStream) -> TokenStream {
...
+                        .unwrap_or(false);
+                    if flatten {
+                        let ident = f.ident.unwrap();
```
mxinden, line 32:
```suggestion
                    let ident = f.ident.unwrap();
                    if flatten {
```

This can be pulled outside of the `if` `else`, right?

derive-encode/src/lib.rs
```diff
@@ -22,22 +22,35 @@ pub fn derive_encode_label_set(input: TokenStream) -> TokenStream {
...
+                        quote! {
+                            self.#ident.encode(encoder)?;
```
mxinden, line 34: The to-be-flattened object needs to be a `struct` and not an `enum`. In other words it needs to implement `EncodeLabelSet` and not `EncodeLabelValue`. Both traits have the same method name, namely `encode`. To prevent a misuse of the `EncodeLabelValue` trait, how about a fully qualified call here?

```suggestion
                            EncodeLabelSet::encode(self.#ident, encoder)?;
```

mxinden (commented):
Great work. Thanks @howardjohn. I have a couple of questions and comments. Overall this looks good to me.

We can release this as a patch release. Would you mind bumping the patch version in `derive-encode/Cargo.toml` and provide a changelog entry in `CHANGELOG.md`?

derive-encode/src/lib.rs
howardjohn: Makes sense. Good idea, thanks!

CHANGELOG.md
```diff
@@ -10,8 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Introduce `Collector` abstraction allowing users to provide additional metrics
   and their description on each scrape. See [PR 82].
+- Introduce a `#[prometheus(flatten)]` attribute which can be used when deriving `EncodeLabelSet`, allowing
+  a nested struct the be flattened during encoding. See [PR 118].
```
mxinden, line 14:
```suggestion
  a nested struct to be flattened during encoding. See [PR 118].
```

CHANGELOG.md
```diff
@@ -10,8 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
```
mxinden, line 15: What do you think of adding your test as an example here?

``` rust
    #[derive(EncodeLabelSet, Hash, Clone, Eq, PartialEq, Debug)]
    struct CommonLabels {
        a: u64,
        b: u64,
    }
    #[derive(EncodeLabelSet, Hash, Clone, Eq, PartialEq, Debug)]
    struct Labels {
        unique: u64,
        #[prometheus(flatten)]
        common: CommonLabels,
    }

    // Would result in `my_metric{a="42",b="42",unique="42"} 42
```

I am sure folks upgrading would greatly appreciate it.

mxinden (commented):
Wonderful. Thanks for the follow-up. One more thing for the changelog. Otherwise this is ready to go. Happy to cut a release right once this is merged.

CHANGELOG.md
howardjohn: Great idea, I updated it

mxinden:
For legal reasons you need to add a `sign-off` line to each of your commits :(, see failing CI check. Mind doing a force push. Sorry for the inconvenience @howardjohn.

howardjohn:
No problem, updated

mxinden (approved):
Thanks for the contribution! Thanks for bearing with me.

mxinden:
[Tagged](https://github.com/prometheus/client_rust/releases/tag/prometheus-client-derive-encode-v0.4.1) and [published](https://crates.io/crates/prometheus-client-derive-encode/0.4.1).

PR #123, opened 2023-01-06
protobuf: Fix label encoding

ackintosh:
Fixes https://github.com/prometheus/client_rust/issues/124

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,8 +36,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   my_metric{a="42",b="42",unique="42"} 42
   ```
 
+### Fixed
+
+- Fix label encoding in protobuf feature. See [PR 123].
+
 [PR 82]: https://github.com/prometheus/client_rust/pull/82
 [PR 118]: https://github.com/prometheus/client_rust/pull/118
+[PR 123]: https://github.com/prometheus/client_rust/pull/123
 
 ## [0.19.0]
 
--- a/src/encoding/protobuf.rs
+++ b/src/encoding/protobuf.rs
@@ -585,11 +585,13 @@ mod tests {
             extract_metric_type(&metric_set)
         );
 
-        // GET, 200
+        // the first label
         let metric = family.metrics.first().unwrap();
         assert_eq!(2, metric.labels.len());
         assert_eq!("method", metric.labels[0].name);
-        assert_eq!("GET", metric.labels[0].value);
+        // The order of the labels is not deterministic so we are testing the value to be either
+        // GET or POST.
+        assert!(vec!["GET", "POST"].contains(&metric.labels[0].value.as_str()));
         assert_eq!("status", metric.labels[1].name);
         assert_eq!("200", metric.labels[1].value);
 
@@ -603,13 +605,12 @@ mod tests {
             _ => panic!("wrong value type"),
         }
 
-        // POST, 200
-        let metric = &family.metrics[1];
-        assert_eq!(2, metric.labels.len());
-        assert_eq!("method", metric.labels[0].name);
-        assert_eq!("POST", metric.labels[0].value);
-        assert_eq!("status", metric.labels[1].name);
-        assert_eq!("200", metric.labels[1].value);
+        // the second label
+        let metric2 = &family.metrics[1];
+        assert_eq!(2, metric2.labels.len());
+        assert!(vec!["GET", "POST"].contains(&metric.labels[0].value.as_str()));
+        assert_eq!("status", metric2.labels[1].name);
+        assert_eq!("200", metric2.labels[1].value);
     }
 
     #[test]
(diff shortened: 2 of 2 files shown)


ackintosh:
I have added a test case to reproduce https://github.com/prometheus/client_rust/issues/124. I was looking into this but couldn't find how to fix. 😓

divagant-martian:
Code is a bit hard to get due to lack of documentation / generic names. Based on the [`text::MetricEncoder`](https://github.com/prometheus/client_rust/blob/9141c118d3c2194c841edca9219dab6d00a19a81/src/encoding/text.rs#L106-L112) field `family_labels`, I assume the [`protobuf::MetricEncoder`](https://github.com/prometheus/client_rust/blob/9141c118d3c2194c841edca9219dab6d00a19a81/src/encoding/protobuf.rs#L109-L113) labels are also meant to be shared labels. 

So if I add this diff
```diff
diff --git a/src/encoding/protobuf.rs b/src/encoding/protobuf.rs
index f40338b..d1f58b2 100644
--- a/src/encoding/protobuf.rs
+++ b/src/encoding/protobuf.rs
@@ -193,6 +193,7 @@ impl<'a> MetricEncoder<'a> {
         &'b mut self,
         label_set: &S,
     ) -> Result<MetricEncoder<'b>, std::fmt::Error> {
+        self.labels.clear();
         label_set.encode(
             LabelSetEncoder {
                 labels: self.labels,
@@ -563,7 +564,7 @@ mod tests {
         family
             .get_or_create(&vec![
                 ("method".to_string(), "GET".to_string()),
-                ("status".to_string(), "200".to_string()),
+                ("status".to_string(), "404".to_string()),
             ])
             .inc();
 
@@ -591,7 +592,7 @@ mod tests {
         assert_eq!("method", metric.labels[0].name);
         assert_eq!("GET", metric.labels[0].value);
         assert_eq!("status", metric.labels[1].name);
-        assert_eq!("200", metric.labels[1].value);
+        assert_eq!("404", metric.labels[1].value);
 
         match extract_metric_point_value(&metric_set) {
             openmetrics_data_model::metric_point::Value::CounterValue(value) => {

```

the test you added gets the correct metric labels, but the test `encoding::protobuf::tests::encode_counter_family_with_prefix_with_label` misses the prefix label.

So I think the problem is that the "family" labels are being populated with normal labels somewhere, can't point to where yet

ackintosh:
I'm not fully sure if I understand current implementation, we should separate labels to e.g. const_labels and family_labels like [text encoding](https://github.com/prometheus/client_rust/blob/9141c118d3c2194c841edca9219dab6d00a19a81/src/encoding/text.rs#L106-L112)? @mxinden How do you think? 🙏

mxinden:
@divagant-martian proposal above, namely to `clear()` the `self.labels` on a call to `encode_family` sounds good to me. This is inline with what the text encoder does, namely to override any existing family labels:

https://github.com/prometheus/client_rust/blob/9141c118d3c2194c841edca9219dab6d00a19a81/src/encoding/text.rs#L210

@ackintosh would you mind applying the suggestion from @divagant-martian?

ackintosh:
I have applied the suggestion but the `encode_counter_family_with_prefix_with_label` test fails. 🤔 

https://github.com/prometheus/client_rust/actions/runs/3958633415/jobs/6780437752#step:9:575
> ---- encoding::protobuf::tests::encode_counter_family_with_prefix_with_label stdout ----
thread 'encoding::protobuf::tests::encode_counter_family_with_prefix_with_label' panicked at 'assertion failed: `(left == right)`
>  left: `3`,
> right: `2`', src/encoding/protobuf.rs:645:9

divagant-martian:
I said this test would fail in my comment

> the test you added gets the correct metric labels, but the test `encoding::protobuf::tests::encode_counter_family_with_prefix_with_label` misses the prefix label.
> 
> So I think the problem is that the "family" labels are being populated with normal labels somewhere, can't point to where yet

So what I think should happen is that this field should be empty in the new test since none of the labels are shared, but populated with the prefix label in the second test. Still don't know where this change should be done tho

ackintosh:
Yeah that is why I said we might want to const_labels and family_labels like text encoding in [my comment](https://github.com/prometheus/client_rust/pull/123#issuecomment-1378044050).

ackintosh:
@mxinden @divagant-martian The bug should have been fixed, please have a look. 🙏

ackintosh:
I have added a doc for `encoding::protobuf::MetricEncoder` for now. 🙂

mxinden:
Thanks for the help here @ackintosh and @divagant-martian!

ackintosh:
@mxinden Could you please cut a new release? I'm thinking of including this fix in rust-libp2p.

divagant-martian:
friendly ping @mxinden for whenever you have the time or if you are waiting on important changes for a release :black_heart:

mxinden:
Sorry for the delay. Tagged and published with https://github.com/prometheus/client_rust/pull/133.

Issue #125, opened 2023-01-10
Undeterministic order in output


lstrojny:
The output or `encode()` is non-deterministic when creating multiple metrics of the same type.

So basically:
```rust
pub fn repro() -> String {
    let mut registry = <prometheus_client::registry::Registry>::default();

    let gauge = Family::<MyLabels, Gauge<f64, AtomicU64>>::default();
    registry.register("test", "help", gauge.clone());

    gauge
        .get_or_create(&MyLabels {
            label: "one".into(),
        })
        .set(0.1);
    gauge
        .get_or_create(&MyLabels {
            label: "two".into(),
        })
        .set(0.2);

    let mut buffer = String::new();

    encode(&mut buffer, &registry).unwrap();

    buffer
}
```

This will sometimes return this:

```
HELP test help.
# TYPE test gauge
test{label="one"} 0.1
test{label="two"} 0.2
# EOF
```

And sometimes this:

```
HELP test help.
# TYPE test gauge
test{label="two"} 0.2
test{label="one"} 0.1
# EOF
```

I would expect the metrics to be somehow sorted deterministically.

mxinden:
I expect this to be due to the use of `HashMap` in `Family`. It could be an option to use a different, ordered datastructure.

Related discussion https://github.com/prometheus/client_golang/issues/483.

> I would expect the metrics to be somehow sorted deterministically.

While I understand the value of sorted output for the sake of debugging, I am not sure it is worth than the (expected) performance hit.

lstrojny:
That makes sense. My use case here is unit testing where I need to assert on a stable output, so I could even make a potential sorting flag dependent on cfg(test). How about adding optional sorting?

mxinden:
> How about adding optional sorting?

Have to give this more thought, though not opposed to it. Might as well always do it in debug mode.

Again, might be worth exploring replacing the `HashMap` and thus getting sorting _for free_.

08d2:
The text encoding of metrics is defined by the spec, and the spec doesn't define any sort order, so callers can't assume any sort order.

cratelyn:
I was looking into this for the same reasons as @lstrojny, and found this thread on the `rust-lang` board: 

https://users.rust-lang.org/t/hashmap-vs-btreemap/13804

> checked just now and BTreeMap is significantly faster than HashMap for many of
> the benchmarks in our `json-benchmark`. (In the table lower is better.)
> 
> | Benchmark | BTreeMap | HashMap |
[...]

PR #128, opened 2023-01-18
Replace HashMap with BTreeMap to sort metrics deterministically

lstrojny:
Addresses #125


08d2:
Is sorting metrics in this way guaranteed by the spec?

lstrojny:
To my knowledge the spec doesn’t say anything about ordering so an implementation is free to do whatever it chooses to do. The goal here is not to implement any particular ordering but instead implement some deterministic behavior to enable e.g. reliable automated tests.

08d2:
If the spec doesn't define deterministic output, then tests can't rely on any output ordering, right? Tests are written not against implementations but against specifications. However, reliable tests can still be achieved, by sorting the output of metric scrapes _within the test_, and evaluating against the result of those sorts.

PR #129, opened 2023-01-19
Timestamp support for const metrics

lstrojny:
@mxinden could you have a look if this is roughly what you had in mind?

### DIFF
--- a/src/encoding/text.rs
+++ b/src/encoding/text.rs
@@ -31,6 +31,7 @@ use crate::registry::{Descriptor, Registry, Unit};
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::fmt::Write;
+use std::time::{SystemTime, UNIX_EPOCH};
 
 /// Encode the metrics registered with the provided [`Registry`] into the
 /// provided [`Write`]r using the OpenMetrics text format.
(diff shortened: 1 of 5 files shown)


lstrojny:
@mxinden any thoughts?

src/encoding/text.rs
```diff
@@ -336,6 +342,15 @@ impl<'a, 'b> MetricEncoder<'a, 'b> {
 
         Ok(())
     }
+
+    fn write_timestamp(&mut self, timestamp: SystemTime) -> Result<(), std::fmt::Error> {
+        if let Ok(time) = timestamp.duration_since(UNIX_EPOCH) {
+            self.writer.write_char(' ')?;
+            return self.writer.write_str(&time.as_millis().to_string());
+        }
```
mxinden, line 350: Should we panic in debug mode?

src/encoding/text.rs
```diff
@@ -336,6 +342,15 @@ impl<'a, 'b> MetricEncoder<'a, 'b> {
 
         Ok(())
     }
+
+    fn write_timestamp(&mut self, timestamp: SystemTime) -> Result<(), std::fmt::Error> {
+        if let Ok(time) = timestamp.duration_since(UNIX_EPOCH) {
+            self.writer.write_char(' ')?;
+            return self.writer.write_str(&time.as_millis().to_string());
```
mxinden, line 349: This library follows the OpenMetrics specification. This should be compatible with recent Prometheus versions (remember to set the right content type).

Please use seconds instead of milliseconds here.

https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#timestamps-1

mxinden (commented):
I am sorry for the late review here.

The direction looks good to me. Great job working through the many layers.

Can you add the protobuf implementation. Also would you mind adding a changelog entry?

PR #130, opened 2023-03-14
Add .get() to Family hashmap wrapper

JaydenElliott:
I have come across a few instances were I would like to check if a particular metric exists in the metric `Family` without creating a new entry if it doesn't. This PR simply exposes the `HashMap::get()` function to allow for `match metric_family.get()` type logic. 

Let me know if there is anything else needed to get this change merged.

### DIFF
--- a/src/metrics/family.rs
+++ b/src/metrics/family.rs
@@ -248,6 +248,23 @@ impl<S: Clone + std::hash::Hash + Eq, M, C: MetricConstructor<M>> Family<S, M, C
         })
     }
 
+    /// Access a metric with the given label set, returning None if one
+    /// does not yet exist.
+    ///
+    /// ```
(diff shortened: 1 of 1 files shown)


mxinden:
Thanks for the pull request.

In order for me to better understand the feature request, mind expanding on your concrete use-cases for this method?

JaydenElliott:
https://github.com/prometheus/client_rust/issues/131#issuecomment-1481962341 describes the use case (allows me to access the metrics of a family to do comparisons against, or use to trigger logic without modifying the underlying data structure).

cratelyn:
this was done in #234, i believe this can be closed now.

mxinden:
You are right. Thanks @cratelyn!

PR #136, opened 2023-04-19
Override derived label value case

drbrain:
By default deriving label values for an enum:

```rust
#[derive(EncodeLabelValue, Hash, Clone, Eq, PartialEq, Debug)]
enum EnumLabel {
    One,
}
```

Results in a label that matches the variant identifier's, here `"One"`.

To allow compatibility with metrics emitted from other processes, or compatibility when switching crates, this PR allows overriding the default case for EncodeLabelValue for an entire enum or for individual enum entries.

```rust
#[derive(EncodeLabelValue, Hash, Clone, Eq, PartialEq, Debug)]
#[prometheus(value_case = "lower")]
enum EnumLabel {
    One,
    Two,
}
```

Would have labels for values `EnumLabel::One` and `EnumLabel::Two` would be `"one"` and `"two"` respectively.

The case can be overridden for a single variant as well:

```rust
#[derive(EncodeLabelValue, Hash, Clone, Eq, PartialEq, Debug)]
enum EnumLabel {
    #[prometheus(lower)]
    One,
    Two,
}
```

Would have labels for values `EnumLabel::One` and `EnumLabel::Two` would be `"one"` and `"Two"` respectively.


derive-encode/src/lib.rs
```diff
@@ -88,22 +88,74 @@ pub fn derive_encode_label_set(input: TokenStream) -> TokenStream {
...
+/// enum Method {
+///     Get,
+///     Put,
+/// }
+/// ```
+///
+/// Will encode to label values "GET" and "PUT" in prometheus metrics.
+///
+/// For variants you can use `#[prometheus(lower)]` or `#[prometheus(upper)]` to set the case for
+/// only that variant.
```
mxinden, line 110: Do you have a concrete use-case where one would upper or lower case a specific variant only?

derive-encode/src/lib.rs
```diff
@@ -88,22 +88,74 @@ pub fn derive_encode_label_set(input: TokenStream) -> TokenStream {
...
+            } else {
+                None
+            }
+        })
+        .unwrap_or_default();
+
     let body = match ast.clone().data {
         syn::Data::Struct(_) => {
-            panic!("Can not derive EncodeLabel for struct.")
+            panic!("Can not derive EncodeLabelValue for struct.")
```
mxinden, line 133: :pray:

derive-encode/src/lib.rs
```diff
@@ -132,6 +184,77 @@ pub fn derive_encode_label_value(input: TokenStream) -> TokenStream {
...
+                    match case.value().as_str() {
+                        "lower" => config.value_case = ValueCase::Lower,
+                        "upper" => config.value_case = ValueCase::Upper,
+                        invalid => {
+                            return Err(syn::Error::new(
+                                case.span(),
+                                format!(
+                                "value case may only be \"lower\" or \"upper\", not \"{invalid}\""
+                            ),
+                            ))
```
mxinden, line 237: To de-duplicate this code with the one above in line 146, how about implementing `FromStr` for `ValueCase`?

mxinden (commented):
I am sorry for the delay here. Well done. Thanks @drbrain!

derive-encode/src/lib.rs
drbrain: I do not

it might be better to remove this and wait for a motivated individual to plumb serde in to label generation which could provide all manner of derive customization

derive-encode/src/lib.rs
drbrain: It will take me some time, but I think I can do this

derive-encode/src/lib.rs
drbrain: Do you want me to remove per-variant case changing?

mxinden: Yes please remove. Thank you.

PR #138, opened 2023-05-09
refactor(family): introduce `ConstFamily` type

mxinden:
Addresses feedback raised by @thomaseizinger in https://github.com/prometheus/client_rust/pull/82/files#r1158205762.

@thomaseizinger this is what we discussed in Brussels. What do you think?

### DIFF
--- a/src/metrics/family.rs
+++ b/src/metrics/family.rs
@@ -327,11 +327,24 @@ where
     }
 }
 
-impl<S: EncodeLabelSet, M: EncodeMetric + TypedMetric, T: Iterator<Item = (S, M)>> EncodeMetric
-    for RefCell<T>
+/// As a [`Family`], but constant, meaning it cannot change once created.
+///
+/// Needed for advanced use-cases, e.g. in combination with [`Collector`](crate::collector::Collector).
+#[derive(Debug, Default)]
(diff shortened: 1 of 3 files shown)


src/metrics/family.rs
```diff
@@ -327,11 +327,24 @@ where
     }
 }
 
-impl<S: EncodeLabelSet, M: EncodeMetric + TypedMetric, T: Iterator<Item = (S, M)>> EncodeMetric
-    for RefCell<T>
+/// As a [`Family`], but constant, meaning it cannot change once created.
+///
+/// Needed for advanced use-cases, e.g. in combination with [`Collector`](crate::collector::Collector).
+#[derive(Debug, Default)]
+pub struct ConstFamily<I>(RefCell<I>);
```
thomaseizinger, line 334: Perhaps use a boxed iterator in here? Then the type parameters would be S and M like for the other families.

mxinden: Thanks for taking a look!

I expect all usages of `ConstFamily` to be short lived, most of them even only within a single parent function scope. Thus I doubt users would ever need to specify `I`, i.e. that it can be inferred by the compiler.

PR #143, opened 2023-05-25
Find the proper bucket before aquiring the lock

vkrasnov:
Currently to find the correct bucket the code has to iterate over the buckets in order under a RWLock. This however takes some time and creates lock contention in heavily threaded access which can be significantly reduced if the correct bucket is found before the lock is aquired. This change creates a copy of the buckets outside the lock and iterates over it instead. Thus the only operations under lock would be the three additions.

### DIFF
--- a/src/metrics/histogram.rs
+++ b/src/metrics/histogram.rs
@@ -35,12 +35,14 @@ use std::sync::Arc;
 // https://github.com/tikv/rust-prometheus/pull/314.
 #[derive(Debug)]
 pub struct Histogram {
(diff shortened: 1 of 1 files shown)


mxinden (commented):
Thank you for the patch. Can you provide some profiling data (e.g. a flamegraph) under production workload, that supports the argument that this is a bottleneck? Also, mind adding a benchmark as a proof?

mxinden:
Closing here due to missing proof. Please open a new pull request once you have data. Thanks for the help!

PR #145, opened 2023-05-28
feat(registry): add `sub_registry_with_labels`

### LINKED ISSUE #144: Allow multiple labels in sub_registry_with_label method
Hey

I have a bunch of counters coming from from processes running on different hosts and I'd like to have them labeled with about 2-3 labels to be able to query them in my dashboards. 

Is there a reason `sub_registry_with_label` accepts a single label? What if I want to add 2-3 labels to a bunch of metrics (globally)?

popadi:
- Added `sub_registry_with_labels` method per discussion in https://github.com/prometheus/client_rust/issues/144
- Updated the tests (which should be refactored or better documented I think, it's a bit hard to figure out the hierarchy or registries)

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -239,11 +239,20 @@ impl Registry {
         &mut self,
         label: (Cow<'static, str>, Cow<'static, str>),
     ) -> &mut Self {
-        let mut labels = self.labels.clone();
-        labels.push(label);
+        self.sub_registry_with_labels(std::iter::once(label))
+    }
+
+    /// Like [`Registry::sub_registry_with_prefix`] but with multiple labels instead.
+    pub fn sub_registry_with_labels(
+        &mut self,
+        labels: impl Iterator<Item = (Cow<'static, str>, Cow<'static, str>)>,
+    ) -> &mut Self {
+        let mut new_labels = self.labels.clone();
+        new_labels.extend(labels);
+
         let sub_registry = Registry {
             prefix: self.prefix.clone(),
-            labels,
+            labels: new_labels,
             ..Default::default()
         };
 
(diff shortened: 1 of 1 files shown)


popadi:
@mxinden any update/feedback on this?

mxinden (commented):
Change looks good to me. Thanks. Though now that we have `sub_registry_with_labelS` we can simply `sub_registry_with_label`, right? What do you think?

``` diff
modified   src/registry.rs
@@ -239,11 +239,20 @@ impl Registry {
         &mut self,
         label: (Cow<'static, str>, Cow<'static, str>),
     ) -> &mut Self {
-        let mut labels = self.labels.clone();
-        labels.push(label);
+        self.sub_registry_with_labels(std::iter::once(label))
+    }
+
+    /// Like [`Registry::sub_registry_with_prefix`] but with multiple labels instead.
+    pub fn sub_registry_with_labels(
+        &mut self,
+        labels: impl Iterator<Item = (Cow<'static, str>, Cow<'static, str>)>,
+    ) -> &mut Self {
+        let mut new_labels = self.labels.clone();
+        new_labels.extend(labels);
+
         let sub_registry = Registry {
             prefix: self.prefix.clone(),
-            labels,
+            labels: new_labels,
             ..Default::default()
         };
 ```

popadi:
> Change looks good to me. Thanks. Though now that we have `sub_registry_with_labelS` we can simply `sub_registry_with_label`, right? What do you think?

Thanks for the suggestion! I updated the code.
Tests might need some extra work (I think individual tests per method would also be useful, alongside the existing one), but I will probably do that in a separate pull request.

mxinden (approved):
Thanks for the follow-ups.

PR #147, opened 2023-05-29
feat(registry): add `with_prefix_and_labels` constructor

### LINKED ISSUE #146: Allow root registry creation with prefix AND label(s)
As the title says, is it okay to add a constructor (`with_prefix_and_labels`; please suggest a better name eventually) that allows the creation of a base/root registry with a prefix and some default labels? Or maybe a way to allow the mutation of the labels of a registry.

I have a bunch of counters coming from from processes running on different hosts and I'd like to have them labeled with about 2-3 base/identification labels to be able to query them in my dashboards properly. Due to some restrictions I am not able to inject these labels directly via kubernetes on Prometheus scrape. At the same time, I see no reason TO NOT allow the creation of a registry with default labels. 

This is a question related a bit to the previously posted issue (https://github.com/prometheus/client_rust/issues/144) for which I submitted a pull request (https://github.com/prometheus/client_rust/pull/145).

popadi:
Added new constructor per discussion in https://github.com/prometheus/client_rust/issues/146.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -75,6 +75,18 @@ impl Registry {
         }
     }
 
+    /// Creates a new default [`Registry`] with the given prefix and labels.
+    pub fn with_prefix_and_labels(
+        prefix: impl Into<String>,
+        labels: impl Iterator<Item = (Cow<'static, str>, Cow<'static, str>)>,
+    ) -> Self {
+        Self {
+            prefix: Some(Prefix(prefix.into())),
+            labels: labels.into_iter().collect(),
+            ..Default::default()
+        }
+    }
+
     /// Register a metric with the [`Registry`].
     ///
     /// Note: In the Open Metrics text exposition format some metric types have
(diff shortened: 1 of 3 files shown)


mxinden (commented):
Put simply there are two ways to create a registry:

1. From scratch e.g. via `Registry::with_prefix` or `<Registry as Default>::default`.
2. From a parent registry e.g. via `Registry::sub_registry_with_prefix`.

What do you think of changing this patch to follow (2) and not (1). It would be a convenience method for folks wanting a sub registry with a prefix and one or more labels. It would still be possible for folks that want (1) via `Registry::default().sub_registry_with_prefix_and_labels(..)`.

popadi:
> What do you think of changing this patch to follow (2) and not (1). It would be a convenience method for folks wanting a sub registry with a prefix and one or more labels. 

Is there any reason there is a separate `Registry::with_prefix` if `Registry::sub_registry_with_prefix` already exist then? The root registry with default labels might be not used that much, but it's definitely a usecase people might have. 

> It would still be possible for folks that want (1) via `Registry::default().sub_registry_with_prefix_and_labels(..)`.

I think we might still want something similar to (1), at least that's my usecase (and that's why I made this pull request in the first place). If I *really* want ALL my metrics to have a prefix AND a bunch of labels, it seems a bit weird to simulate the root registry using a sub-registry.

popadi:
@mxinden any update/feedback on this?

mxinden (commented):
Sorry for the delay here. Thank you for expanding on your use-case. Agreed that we should support it.

For the sake of consistency, what do you think about also adding `Registry::with_labels`?

Also can you bump the crate version in `Cargo.toml` to `v0.21.2` and add a changelog entry in `CHANGELOG.md`?

popadi:
> Sorry for the delay here. Thank you for expanding on your use-case. Agreed that we should support it.
> 
> For the sake of consistency, what do you think about also adding `Registry::with_labels`?
> 
> Also can you bump the crate version in `Cargo.toml` to `v0.21.2` and add a changelog entry in `CHANGELOG.md`?

Done, added the extra method + updated changelog and bumped version.
Thanks for the feedback

src/registry.rs
```diff
@@ -518,6 +540,54 @@ mod tests {
...
+
+        assert_eq!(
+            Some((prefix.to_string() + "_" + counter_name, vec![])),
+            registry
+                .iter_metrics()
+                .map(|(desc, _)| (desc.name.clone(), desc.labels.clone()))
+                .next()
+        );
+
+        // test with_labels constructor
```
mxinden, line 565:
```suggestion
        // test with_labels constructor
```

src/registry.rs
```diff
@@ -518,6 +540,54 @@ mod tests {
...
 
+    #[test]
+    fn constructors() {
+        let counter_name = "test_counter";
+        let prefix = "test_prefix";
+        let labels = vec![
+            (Cow::Borrowed("global_label_1"), Cow::Borrowed("value_1")),
+            (Cow::Borrowed("global_label_1"), Cow::Borrowed("value_2")),
+        ];
+
```
mxinden, line 551:
```suggestion

```

mxinden (approved):
Thank you for the continued work!

mxinden:
[Published](https://crates.io/crates/prometheus-client/0.21.2) and [tagged](https://github.com/prometheus/client_rust/releases/tag/v0.21.2) `prometheus-client` `v0.21.2`.

PR #149, opened 2023-06-04
refactor(collector): have Registry:encode do the encoding 

mxinden:
Let's face it, the `Collector` trait I designed in https://github.com/prometheus/client_rust/pull/82 is too complex.

Instead of having `Registry` provide access to its inner `Metric`s, have
`Registry` encode the `Metric`s directly via `Registry::encode`. In the same
vain, have `Collector::encode` encode the collected `Metric`s directly.

This allows for a much simpler `Collector` trait, removing all complexity due to
providing access to a `Metric` externally.

``` rust
/// The [`Collector`] abstraction allows users to provide additional metrics and
/// their description on each scrape.
///
/// An example use-case is an exporter that retrieves a set of operating system metrics
/// ad-hoc on each scrape.
///
/// Register a [`Collector`] with a [`Registry`](crate::registry::Registry) via
/// [`Registry::register_collector`](crate::registry::Registry::register_collector).
///
/// ```
/// # use prometheus_client::metrics::counter::ConstCounter;
/// # use prometheus_client::collector::Collector;
/// # use prometheus_client::encoding::{DescriptorEncoder, EncodeMetric};
/// #
/// #[derive(Debug)]
/// struct MyCollector {}
///
/// impl Collector for MyCollector {
///     fn encode(&self, mut encoder: DescriptorEncoder) -> Result<(), std::fmt::Error> {
///         let counter = ConstCounter::new(42);
///         let metric_encoder = encoder.encode_descriptor(
///             "my_counter",
///             "some help",
///             None,
///             counter.metric_type(),
///         )?;
///         counter.encode(metric_encoder)?;
///         Ok(())
///     }
/// }
/// ```
pub trait Collector: std::fmt::Debug + Send + Sync + 'static {
    /// Once the [`Collector`] is registered, this method is called on each scrape.
    fn encode<'a>(&'a self, encoder: DescriptorEncoder) -> Result<(), std::fmt::Error>;
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.22.0] - unreleased
+
+### Changed
+
+- Simplify `Collector` `trait` by enabling `Collector::collect` to encode metrics directly with a `DescriptorEncoder`.
+  See [PR 149] for details.
+
+[PR 149]: https://github.com/prometheus/client_rust/pull/149
+
 ## [0.21.2]
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ prost-types = { version = "0.11.0", optional = true }
 
 [dev-dependencies]
 async-std = { version = "1", features = ["attributes"] }
-criterion = "0.4"
+criterion = "0.5"
 http-types = "2"
 pyo3 = "0.18"
 quickcheck = "1"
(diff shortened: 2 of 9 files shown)


mxinden:
@thomaseizinger based on your comments in https://github.com/prometheus/client_rust/pull/82#discussion_r1158205762 I think you will like this change.

thomaseizinger:
I think I was able to simplify this a bit in https://github.com/mxinden/client_rust/pull/2.

gurinderu:
@mxinden looks like there is no way to add custom labels with the new API. Probably we should add an additional encode_descriptor function with optional labels. WDYT?

mxinden:
Does the `encode_family` method solve your use-case? Example:

https://github.com/libp2p/rust-libp2p/blob/cddc306978a7cca705de5ce10ec5e16e8b732084/misc/metrics/src/identify.rs#L211-L223

PR #162, opened 2023-09-05
chore: update prost (prost-types & prost-build) from 0.11.0 to 0.12.0

cheshirskycode:
It works only with batch update

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,8 @@ dtoa = "1.0"
 itoa = "1.0"
 parking_lot = "0.12"
 prometheus-client-derive-encode = { version = "0.4.1", path = "derive-encode" }
-prost = { version = "0.11.0", optional = true }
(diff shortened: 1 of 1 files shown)


mxinden:
Thank you for looking into this. I am guessing the failure is related to https://github.com/prometheus/client_rust/pull/98/.

Long run I would like to move away from `prost` towards `quick-protobuf`. We had good experience with that in https://github.com/libp2p/rust-libp2p/issues/3024.

@cheshirskycode what do you think? Would you be interested in contributing the above move from `prost` to `quick-protobuf`?

cheshirskycode:
@mxinden, yes, it is interesting. I'm in.

mxinden:
@cheshirskycode any updates here? Anything I can help with?

krisztianfekete:
We are now on 0.14.x

PR #170, opened 2023-10-24
chore(derive): include license files

decathorpe:
I am packaging the prometheus-client and prometheus-client-derive-encode crates for Fedora Linux. I noticed that the latter crate does not contain license files as it is published to crates.io.

However, both the terms of the MIT and Apache-2.0 licenses require that any (re)distributed sources contain a copy of the original license text. This is potentially problematic for the crates as published on crates.io, but definitely problematic for packaging them on Linux distributions.

This PR includes symbolic links to both license files in the derive-encode directory, which should result in "cargo package" / "cargo publish" including a copy of the license files when publishing to crates.io. This is assuming that the publishing workflow is run on an operating system that supports symbolic links - if publishing happens on Windows, full copies of the license texts might be required instead of a symbolic link.

### DIFF
--- a/derive-encode/LICENSE-APACHE
+++ b/derive-encode/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+   "License" shall mean the terms and conditions for use, reproduction,
+   and distribution as defined by Sections 1 through 9 of this document.
+
+   "Licensor" shall mean the copyright owner or entity authorized by
+   the copyright owner that is granting the License.
+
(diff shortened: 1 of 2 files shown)


mxinden:
Thank you for the patch and thank you for packaging `prometheus-client`. I suggest copying the files for the sake of simplicity. Also, CI is currently failing on the DCO step.

decathorpe:
Thanks for your feedback - I've changed the PR to use copies instead of symlinks.

As for the DCO, it looks like I need to add a "Signed-off-by" line to the commits. What am I signing off on by doing that, exactly? I couldn't find any documentation for that.

mxinden:
Fair question. You can find more details here https://probot.github.io/apps/dco/.

As far as I know this is a legal requirement forced upon the project. Personally I find it very annoying. That said, it might be more useful than I think, i.e. I might be misjudging the legal risk the project might face without.

mxinden (approved):
Cool. Thanks.

PR #173, opened 2023-10-28
feat(encoding): add impls for `Encode*Value`

### LINKED ISSUE #172: mips: EncodeCounterValue implementation missing for u32?
Code that compiles fine for ARMv7 and AMD64, fails when compiling for MIPS (`mipsel-unknown-linux-musl`):

```
...
the trait `EncodeCounterValue` is not implemented for `u32`
   |
   = help: the following other types implement trait `EncodeCounterValue`:
             u64
             f64
   = note: required for `prometheus_client::metrics::counter::Counter` to implement `EncodeMetric`
   = note: 1 redundant requirement hidden
...
```

It looks like that `mips` targets are using `u32` instead of `u64` for counter values:

https://github.com/prometheus/client_rust/blob/master/src/metrics/counter.rs#L51

And indeed there is no `EncodeCounterValue` implementation for `u32`:

https://github.com/prometheus/client_rust/blob/master/src/encoding.rs#L537

**Edit:** This issue might apply to other types as well, e.g. `Gauge`. I have not tested this.

brocaar:
With these changes, I no longer get the error:

> trait `EncodeCounterValue` is not implemented for `u32`

when compiling to mips.

I have also added additional trait implementations for the other types. The reason to cast from 32 to 64 bit values is that in the end, it is casted to 64 bit values anyway, because these are the types that must be used: https://github.com/prometheus/client_rust/blob/master/src/encoding/proto/openmetrics_data_model.proto. This way, there is no need to implement `encoder.encode_u32` ... methods.

I believe this is not catched by the tests, because the library itself compiles fine to mips, until you start implementing it. Unfortunately the cross-compile check does not run the tests. I did a quick test changing the `cargo check ...` to `cargo check --target=${{ matrix.target }} --tests`, but this will return errors because the dev-dependency `pyo3` a bit more than just the Rust toolchain for mips is required.

Fixes #172.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -500,6 +500,18 @@ impl EncodeGaugeValue for f64 {
     }
 }
 
+impl EncodeGaugeValue for i32 {
+    fn encode(&self, encoder: &mut GaugeValueEncoder) -> Result<(), std::fmt::Error> {
+        encoder.encode_i64(*self as i64)
+    }
+}
+
+impl EncodeGaugeValue for f32 {
+    fn encode(&self, encoder: &mut GaugeValueEncoder) -> Result<(), std::fmt::Error> {
+        encoder.encode_f64(*self as f64)
+    }
+}
+
 /// Encoder for a gauge value.
 #[derive(Debug)]
 pub struct GaugeValueEncoder<'a>(GaugeValueEncoderInner<'a>);
@@ -552,6 +564,18 @@ impl EncodeCounterValue for f64 {
     }
 }
 
+impl EncodeCounterValue for u32 {
+    fn encode(&self, encoder: &mut CounterValueEncoder) -> Result<(), std::fmt::Error> {
+        encoder.encode_u64(*self as u64)
+    }
+}
+
+impl EncodeCounterValue for f32 {
+    fn encode(&self, encoder: &mut CounterValueEncoder) -> Result<(), std::fmt::Error> {
+        encoder.encode_f64(*self as f64)
+    }
+}
+
 /// Encoder for a counter value.
 #[derive(Debug)]
 pub struct CounterValueEncoder<'a>(CounterValueEncoderInner<'a>);
@@ -586,6 +610,18 @@ impl EncodeExemplarValue for f64 {
 }
 
 impl EncodeExemplarValue for u64 {
+    fn encode(&self, mut encoder: ExemplarValueEncoder) -> Result<(), std::fmt::Error> {
+        encoder.encode(*self as f64)
+    }
+}
+
+impl EncodeExemplarValue for f32 {
+    fn encode(&self, mut encoder: ExemplarValueEncoder) -> Result<(), std::fmt::Error> {
+        encoder.encode(*self as f64)
+    }
+}
+
+impl EncodeExemplarValue for u32 {
     fn encode(&self, mut encoder: ExemplarValueEncoder) -> Result<(), std::fmt::Error> {
         encoder.encode(*self as f64)
     }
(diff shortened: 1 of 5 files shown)


brocaar:
As a reference, this is the error I get when using `cargo check ... --tests`:

```
error: failed to run custom build command for `pyo3-ffi v0.20.0`

Caused by:
  process didn't exit successfully: `/home/runner/work/client_rust/client_rust/target/debug/build/pyo3-ffi-4fe3cdc12a0f9dda/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=PYO3_CROSS
  cargo:rerun-if-env-changed=PYO3_CROSS_LIB_DIR
  cargo:rerun-if-env-changed=PYO3_CROSS_PYTHON_VERSION
  cargo:rerun-if-env-changed=PYO3_CROSS_PYTHON_IMPLEMENTATION
  cargo:rerun-if-env-changed=PYO3_NO_PYTHON

  --- stderr
  error: PYO3_CROSS_PYTHON_VERSION or an abi3-py3* feature must be specified when cross-compiling and PYO3_CROSS_LIB_DIR is not set.
warning: build failed, waiting for other jobs to finish...
error: failed to run custom build command for `zstd-sys v2.0.9+zstd.1.5.5`
```

src/encoding.rs
```diff
@@ -552,6 +564,24 @@ impl EncodeCounterValue for f64 {
...
 
+impl EncodeCounterValue for u32 {
+    fn encode(&self, encoder: &mut CounterValueEncoder) -> Result<(), std::fmt::Error> {
+        encoder.encode_u64(*self as u64)
+    }
+}
+
+impl EncodeCounterValue for i32 {
+    fn encode(&self, encoder: &mut CounterValueEncoder) -> Result<(), std::fmt::Error> {
+        encoder.encode_u64(*self as u64)
```
mxinden, line 575:
```suggestion
        encoder.encode_i64(*self as u64)
```

src/encoding.rs
```diff
@@ -552,6 +564,24 @@ impl EncodeCounterValue for f64 {
     }
 }
 
+impl EncodeCounterValue for u32 {
+    fn encode(&self, encoder: &mut CounterValueEncoder) -> Result<(), std::fmt::Error> {
+        encoder.encode_u64(*self as u64)
+    }
+}
+
+impl EncodeCounterValue for i32 {
```
mxinden, line 573: A counter can never be negative. Why implement `EncodeCounterValue` for `i32`?

mxinden (commented):
I am sorry for the delay here.

Thank you for debugging and sending a patch!

I would appreciate any additions to the CI to prevent this from happening in the future.

src/encoding.rs
brocaar: Are you sure? https://github.com/prometheus/client_rust/blob/master/src/encoding.rs#L571

```
impl<'a> CounterValueEncoder<'a> {
    fn encode_f64(&mut self, v: f64) -> Result<(), std::fmt::Error> {
        for_both_mut!(self, CounterValueEncoderInner, e, e.encode_f64(v))
    }

    fn encode_u64(&mut self, v: u64) -> Result<(), std::fmt::Error> {
        for_both_mut!(self, CounterValueEncoderInner, e, e.encode_u64(v))
    }
}
```

src/encoding.rs
brocaar: Fixed.

brocaar:
@mxinden I'm looking forward to your feedback on this :-)

brocaar:
Hi @mxinden I'm still looking forward to your feedback :-)

mxinden (approved):
Thank you for the work and once again, sorry for the delay.

PR #175, opened 2023-10-31
fix: correctly handle empty family labels

mxinden:


src/encoding/text.rs
```diff
@@ -742,6 +746,24 @@ mod tests {
         parse_with_python_client(encoded);
     }
 
+    #[test]
+    fn encode_histogram_family_with_empty_family_labels() {
+        let mut registry = Registry::default();
+        let family =
+            Family::new_with_constructor(|| Histogram::new(exponential_buckets(1.0, 2.0, 10)));
+        registry.register("my_histogram", "My histogram", family.clone());
```
mxinden, line 754: Just to make sure we are on the same page, you don't need to wrap a `Histogram` in a `Family`.

```suggestion
        let histogram = Histogram::new(exponential_buckets(1.0, 2.0, 10));
        registry.register("my_histogram", "My histogram", historgam.clone());
```

This should fix the encoding failure, correct?

src/encoding/text.rs
```diff
@@ -742,6 +746,24 @@ mod tests {
...
+
+        family
+            .get_or_create(&())
```
mxinden, line 757: In what use-case would you want a `Family` with a `()` as a label set?

To remove this footgun, shall we remove the `EncodeLabelSet` implementation for `()`?

https://github.com/prometheus/client_rust/blob/23c31bdacdc2445b0b85e79684a6d2534330fc29/src/encoding.rs#L352-L356

mxinden (commented):
Thank you @tylerlevine for reporting and providing a patch!

src/encoding/text.rs
tylerlevine: Ah - this makes more sense now. I didn't realize that the `Family` wrapper was optional. Using the histogram directly indeed does fix the encoding failure.

src/encoding/text.rs
tylerlevine: Now that I understand that the `Family` wrapper is optional, I agree that using `()` as the label set doesn't make much sense.

However, when I originally ran into this issue I was using an empty struct, which has similar behavior. I was intending to go back and add any needed labels to the struct later but noticed this issue before I could do that. I switched to `()` when I was simplifying my code for posting here.

Removing the `EncodeLabelSet` implementation from `()` seems like a good start, but ideally a fix would cover the empty struct case as well.

src/encoding/text.rs
```diff
@@ -742,6 +747,51 @@ mod tests {
...
+
+        let mut encoded = String::new();
+
+        encode(&mut encoded, &registry).unwrap();
+
+        parse_with_python_client(encoded);
+    }
+
+    #[test]
+    fn encode_histogram_family_with_empty_struct_family_labels() {
```
tylerlevine, line 769: Added this second test to give a more realistic use case of empty family labels.

src/encoding/text.rs
tylerlevine: I took a look at removing the `EncodeLabelSet` implementation for `()`, but it looks like that impl is relied on in at least these four places:

https://github.com/prometheus/client_rust/blob/1f3923a7a53db30081a11d1741a1c6e50ae08111/src/encoding/text.rs#L189

https://github.com/prometheus/client_rust/blob/1f3923a7a53db30081a11d1741a1c6e50ae08111/src/encoding/text.rs#L213

https://github.com/prometheus/client_rust/blob/1f3923a7a53db30081a11d1741a1c6e50ae08111/src/encoding/text.rs#L269

https://github.com/prometheus/client_rust/blob/1f3923a7a53db30081a11d1741a1c6e50ae08111/src/encoding/text.rs#L276

mxinden: I would assume that we can replace `()` with a private uninhabited type that implements `EncodeLabelSet`. Something along the lines of:

``` rust
enum Never {}

impl EncodeLabelSet for Never {
  // ...
}

// ...

// Then in the above cases:
self.encode_labels::<Never>(None)?;
```

Want to give this a shot @tylerlevine ?

tylerlevine: Good idea. Yes, let me give it a try.

mxinden: Friendly ping @tylerlevine. What do you think of the suggestion above?

tylerlevine: @mxinden Looks like that works fine. I cleaned up a bit and updated the PR.

src/encoding/text.rs
```diff
@@ -375,11 +375,15 @@ impl<'a> MetricEncoder<'a> {
...
+            labels.encode(LabelSetEncoder::new(&mut string_writer).into())?;
+
+            if !string_writer.is_empty() {
+                if !self.const_labels.is_empty() || additional_labels.is_some() {
+                    self.writer.write_str(",")?;
+                }
+                self.writer.write_str(string_writer.as_str())?;
             }
-
-            labels.encode(LabelSetEncoder::new(self.writer).into())?;
```
mxinden, line 382: Sorry for only raising this now.

This change is not ideal, as it doubles the allocations required for `self.family_labels`, i.e. adds unnecessary allocations to the happy path.

I see multiple ways to still remove the bug (i.e. an invalid dangling `,`) without the allocation:

1. Fail to derive `EncodeLabelSet` on an empty `struct`. This is a partial fix only as one could still go through e.g. a `Vec<(String, String)>`.
2. Wrap `self.writer` with a thin wrapper, that simply slips in a `,` in case any of its `std::fmt::Write` methods are called. In other words that adds a `,` at very first in case `family_labels` actually attempts to write anything.

Does the above make sense? I am in favor of (2). What do you think?

tylerlevine: Yes, that makes perfect sense to me - thanks for the suggestions. Apologies for my inactivity on this PR, but I'll go for option 2 once I get some time to work on this again. The approach seems pretty straightforward to me.

mxinden:
Friendly ping @tylerlevine.

tylerlevine:
> Friendly ping @tylerlevine.

Thanks for the ping @mxinden - I've updated the PR to implement suggestion 2 from your earlier comment.

mxinden (commented):
Looks good to me. Thank you!

Just missing a changelog entry, otherwise this is good to go.

mxinden:
Merged through https://github.com/prometheus/client_rust/pull/224. Thank you @tylerlevine for the work!

PR #180, opened 2023-11-22
feat(family) Add `with_labels` method to the metrics Family

alekseysidorov:
Added a new method `with_labels` to `Family`

Unlike the `Family::get_or_create` method, this one takes a closure that runs with the context of a counter. This approach avoids label set cloning and better hides an implementation details.

## Example

```rust
let family = Family::<Vec<(String, String)>, Counter>::default();
// Similar to the `get_or_create`
family.with_labels(vec![("method".to_string(), "GET".to_string())], |counter| {
    counter.inc()
});
assert_eq!(
    2,
    family.with_labels(vec![("method".to_string(), "GET".to_string())], |counter| {
        counter.get()
    })
);
```

### DIFF
--- a/benches/family.rs
+++ b/benches/family.rs
@@ -1,54 +1,138 @@
+use std::time::Instant;
+
 use criterion::{criterion_group, criterion_main, Criterion};
 use prometheus_client::metrics::counter::Counter;
 use prometheus_client::metrics::family::Family;
 
 pub fn family(c: &mut Criterion) {
-    c.bench_function("counter family with Vec<(String, String)> label set", |b| {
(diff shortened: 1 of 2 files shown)


mxinden (commented):
Sorry for the delayed review.

> This approach avoids label set cloning

But it requires an owned label set, thus requiring the caller to clone on every call. How is that better to a reference that is only cloned on demand, i.e. when there is no metric for the given label set yet.

> better hides an implementation details.

Which implementation detail is this hiding?

mxinden:
Closing here since this is missing further reasoning why this is worth merging. Please ping me in case I am missing something.

PR #185, opened 2023-12-03
More benches

MOZGIII:
This PR just adds a few more benches that I found useful

### DIFF
--- a/benches/baseline.rs
+++ b/benches/baseline.rs
@@ -1,0 +1,24 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use prometheus_client::metrics::counter::Counter;
+use prometheus_client::metrics::family::Family;
+
+pub fn baseline(c: &mut Criterion) {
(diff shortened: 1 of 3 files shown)


benches/baseline.rs
```diff
@@ -0,0 +1,24 @@
...
+use prometheus_client::metrics::family::Family;
+
+pub fn baseline(c: &mut Criterion) {
+    c.bench_function("counter", |b| {
+        let counter: Counter = Counter::default();
+
+        b.iter(|| {
+            counter.inc();
+        })
+    });
```
mxinden, line 12: This should just boil down to however fast `AtomicU64::fetch_add` is. That said, won't hurt to have it here.

mxinden (approved):
Thank you for the additions. Anything interesting you found?

benches/baseline.rs
MOZGIII: Yep, seemed like a good baseline to me

PR #189, opened 2024-02-28
fix(metrics): :impl Atomic<u64> for AtomicU64

koushiro:
fix the error when using Gauge<u64, AtomicU64> type, revert the `impl Atomic<u64> fro AtomicU64` from #105

### DIFF
--- a/src/metrics/gauge.rs
+++ b/src/metrics/gauge.rs
@@ -188,6 +188,33 @@ impl Atomic<i32> for AtomicI32 {
 }
 
 #[cfg(not(any(target_arch = "mips", target_arch = "powerpc")))]
+impl Atomic<u64> for AtomicU64 {
+    fn inc(&self) -> u64 {
+        self.inc_by(1)
+    }
+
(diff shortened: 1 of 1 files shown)


mxinden:
If I recall correctly, I removed this implementation, given that the protobuf definition does not support `u64`, but only `i64`. With your patch, how would one encode a `u64::MAX` in a protobuf?

https://github.com/OpenObservability/OpenMetrics/blob/1386544931307dff279688f332890c31b6c5de36/proto/openmetrics_data_model.proto#L100-L107

koushiro:
Ok, If the protobuf definition does not support `u64`, why not add `impl Atomic<u32> for AtomicU32`, at least we could use `Gauge<u32, AtomicU32>`

mxinden:
> Ok, If the protobuf definition does not support `u64`, why not add `impl Atomic<u32> for AtomicU32`, at least we could use `Gauge<u32, AtomicU32>`

Sorry for the late reply. I am fine with adding `u32` to `Gauge` @koushiro.

PR #191, opened 2024-03-06
feat(metrics): support `Gauge<u32, AtomicU32>` type

koushiro:
support `Gauge<u32, AtomicU32>` metric

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.22.2]
+
+### Added
+
+- Added `Gauge<u32, AtomicU32>` implementation.
+  See [PR 191].
+
+[PR 191]: https://github.com/prometheus/client_rust/pull/191
+
 ## [0.22.1]
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus-client"
-version = "0.22.1"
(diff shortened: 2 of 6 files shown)


koushiro:
@mxinden PTAL

mxinden (commented):
Cool. Thanks. Looks good to me.

Can you bump the crate version in the root's `Cargo.toml` to `v0.22.2` and add a `CHANGELOG.md` entry? Once this pull request merged, I can then cut a new release right away.

mxinden:
The following test:

``` rust
    #[test]
    fn encode_gauge() {
        let gauge = Gauge::<u32, AtomicU32>::default();
        registry.register("my_gauge", "My gauge", gauge);

        let mut encoded = String::new();

        encode(&mut encoded, &registry).unwrap();

        parse_with_python_client(encoded);
    }
```

Would fail with:

```
   --> src/encoding/text.rs:635:51
    |
635 |         registry.register("my_gauge", "My gauge", gauge);
    |                  --------                         ^^^^^ the trait `EncodeGaugeValue` is not implemented for `u32`
    |                  |
    |                  required by a bound introduced by this call
```

Mind adding an implementation of `EncodeGaugeValue` for `u32`? You can cast into an `i64` and then use the existing implementation for `i64`.

Please also extend the `encode_gauge` test in `text.rs` to register both a default `Gauge` and a `Gauge<u32, AtomicU32` to make sure we don't break this in the future.

Sorry for the additional work. Thank you for the help.

koushiro:
Thanks for your feedback, I have changed encoding related stuff and added a simple test case to `encode_gauge`.
@mxinden PTAL again

mxinden (approved):
Well done!

mxinden:
[Published](https://crates.io/crates/prometheus-client/0.22.2) and [tagged](https://github.com/prometheus/client_rust/releases/tag/v0.22.2).

PR #192, opened 2024-03-28
examples/actix-web.rs: Default enable compress middleware.

jpds:


mxinden:
Thank you for the proposal.

I still remember compression not being a silver bullet. E.g. see https://github.com/kubernetes/kube-state-metrics/issues/498.

Is this the default in other client libraries? Have you done any measurements with and without?

jpds:
This is the default in the Go library: https://github.com/prometheus/client_golang/blob/93cf5d4f5f78058ef0e5fe5fbf597de2084c64fd/prometheus/promhttp/http.go#L172-L175 - I haven't done any measurements yet.

mxinden (commented):
Thank you for the link.

This looks good to me. Can you please fix the `cargo fmt` failure?

mxinden (approved):
:pray:

PR #196, opened 2024-04-29
Prototype of metrics expiry

howardjohn:


howardjohn:
Ok I have fixed this up and added documentation. Should be good to go from my perspective.

PR #199, opened 2024-05-28
Add clear/reset methods for histograms and counters

zakcutner:
Thanks for this great project! I noticed that [families have a `clear` method](https://docs.rs/prometheus-client/0.22.2/prometheus_client/metrics/family/struct.Family.html#method.clear), but nothing like this exists for histograms and counters. I figured out that you can get around this for counters with `.inner().store(0, Ordering::Relaxed)` (although using `.inner()` does not feel so nice to me), but I couldn't find a way to do this for histograms.

I've found that clearing/resetting metrics is particularly useful for testing, to avoid needing to compare each metric to the value it was at the beginning of the test. Therefore, I've added methods to clear/reset histograms and counters in this PR. I'm also happy to first create an issue if some discussion on adding these methods is needed.

### DIFF
--- a/src/encoding/protobuf.rs
+++ b/src/encoding/protobuf.rs
@@ -739,6 +739,35 @@ mod tests {
                 );
                 assert_eq!(1, value.count);
                 assert_eq!(11, value.buckets.len());
+                assert_eq!(1u64, value.buckets.iter().map(|bucket| bucket.count).sum());
+            }
+            _ => panic!("wrong value type"),
+        }
+
+        histogram.clear();
+
+        let metric_set = encode(&registry).unwrap();
+
+        let family = metric_set.metric_families.first().unwrap();
+        assert_eq!("my_histogram", family.name);
+        assert_eq!("My histogram.", family.help);
+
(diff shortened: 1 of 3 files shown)


mxinden:
Hi @zakcutner, thank you for your contribution!



> I've found that clearing/resetting metrics is particularly useful for testing, to avoid needing to compare each metric to the value it was at the beginning of the test. Therefore, I've added methods to clear/reset histograms and counters in this PR. I'm also happy to first create an issue if some discussion on adding these methods is needed.

Is there a use-case for this outside of testing?

Are you aware of any Prometheus client libraries in other languages that offer these methods?

zakcutner:
Thanks for getting back to me @mxinden!

> Is there a use-case for this outside of testing?

Not that I've encountered 🙂 

> Are you aware of any Prometheus client libraries in other languages that offer these methods?

I'm only familiar with the official client library for Golang, and it doesn't look like that one has clear/reset methods. It also doesn't seem to have a clear method for families, like this library already does.

Issue #201, opened 2024-06-25
Some 32 bit platforms are not supported


navaati:
I’m trying to use this library on a 32-bit platform (an ESP32-C3 chip, RISC-V based, has alloc and std).

However when building I get this error:
```console
$ cargo check
    Checking prometheus-client v0.22.2
error[E0432]: unresolved import `std::sync::atomic::AtomicU64`
  --> /home/navaati/.cargo/registry/src/index.crates.io-6f17d22bba15001f/prometheus-client-0.22.2/src/metrics/counter.rs:10:5
   |
10 | use std::sync::atomic::AtomicU64;
   |     ^^^^^^^^^^^^^^^^^^^---------
   |     |                  |
   |     |                  help: a similar name exists in the module: `AtomicU32`
   |     no `AtomicU64` in `sync::atomic`

error[E0432]: unresolved import `std::sync::atomic::AtomicU64`
  --> /home/navaati/.cargo/registry/src/index.crates.io-6f17d22bba15001f/prometheus-client-0.22.2/src/metrics/exemplar.rs:17:5
   |
17 | use std::sync::atomic::AtomicU64;
   |     ^^^^^^^^^^^^^^^^^^^---------
   |     |                  |
   |     |                  help: a similar name exists in the module: `AtomicU32`
   |     no `AtomicU64` in `sync::atomic`

error[E0432]: unresolved imports `std::sync::atomic::AtomicI64`, `std::sync::atomic::AtomicU64`
  --> /home/navaati/.cargo/registry/src/index.crates.io-6f17d22bba15001f/prometheus-client-0.22.2/src/metrics/gauge.rs:11:25
   |
11 | use std::sync::atomic::{AtomicI64, AtomicU64};
   |                         ^^^^^^^^^  ^^^^^^^^^ no `AtomicU64` in `sync::atomic`
   |                         |
   |                         no `AtomicI64` in `sync::atomic`
```

My platform doesn’t support 64 bit atomics. Would there be a way to drop down to 32 bit atomics, perhaps with config detections, or a feature flag ?

navaati:
Ah ! I realize it is actually supported, just for certain listed platforms: https://github.com/prometheus/client_rust/blob/bf196d77824294cce6bf996f79cef25f7aa82c74/src/metrics/counter.rs#L9-L10

I’ll make a PR for ESPs.

navaati:
Related previous bug and PR: #37 #42 .

navaati:
[...]

PR #202, opened 2024-06-25
chore(portability): Support 32 bit RISC-V

navaati:
Adds `target_arch = "riscv32"` to the list of AtomicU64 exclusions.

This is a narrow fix for #201.


navaati:
Abandoning this as #203 is clearly superior and more general.

PR #203, opened 2024-06-25
chore(portability): Support all platforms without 64 bit atomics

navaati:
Notably, RISC-V 32 for ESP32-C3 chips.

### DIFF
--- a/src/metrics/counter.rs
+++ b/src/metrics/counter.rs
@@ -6,7 +6,7 @@ use crate::encoding::{EncodeMetric, MetricEncoder};
 
 use super::{MetricType, TypedMetric};
 use std::marker::PhantomData;
-#[cfg(not(any(target_arch = "mips", target_arch = "powerpc")))]
+#[cfg(target_has_atomic = "64")]
 use std::sync::atomic::AtomicU64;
 use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::Arc;
@@ -40,7 +40,7 @@ use std::sync::Arc;
 /// counter.inc();
 /// let _value: f64 = counter.get();
 /// ```
-#[cfg(not(any(target_arch = "mips", target_arch = "powerpc")))]
+#[cfg(target_has_atomic = "64")]
 #[derive(Debug)]
 pub struct Counter<N = u64, A = AtomicU64> {
     value: Arc<A>,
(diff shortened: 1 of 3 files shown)


mxinden (commented):
`target_has_atomic` is neat! Thank you @navaati. Good work.

Do you have experience cross-compiling Rust for RISC-V? If so, want to add another `target` to the matrix here within this pull request?

https://github.com/prometheus/client_rust/blob/bf196d77824294cce6bf996f79cef25f7aa82c74/.github/workflows/rust.yml#L160-L178

navaati:
Ah, yes. It very much depends on the support from that Github Action you use, which I’m not super familiar with.

The target I actually use, `riscv32imc-esp-espidf`, is Tier 3 and I’m not sure it’ll be supported (among other things, it requires building the std lib), the closest would be `riscv32imc-unknown-none-elf` but it is a no-std target so it makes no sense here…

I’ll put out a commit with `riscv32imc-esp-espidf` in the list and we’ll see how it goes, can you allow the CI to run once it’s here ?

mxinden:
> I’ll put out a commit with riscv32imc-esp-espidf in the list and we’ll see how it goes, can you allow the CI to run once it’s here ?

Yep. Will do. Thanks.

mxinden:
Also note, you will need to sign off your commits, see failing CI step.

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Support all platforms lacking 64 bit atomics.
```
mxinden, line 11:
```suggestion
- Support all platforms with 32 bit atomics lacking 64 bit atomics.
```

Assuming there are non-u32 atomics platforms as well.

mxinden:
Just triggered a CI run.

CHANGELOG.md
navaati: Yeaaaah, I kinda implicitly left those out of the whole equation because you need _something_, but I can clarify the changelog !

navaati:
> Just triggered a CI run.

Thanks. And it fails, as I feared the target needs the rust-std component and it’s nightly only… But you have powerpc and wasm32 which are 32-bit platforms, so it should be ok.

Should I just drop that CI commit and we go as is ?

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
```suggestion
## [0.22.3] - unreleased
```

To make sure we know the below is not a breaking change.

mxinden (approved):
:pray:

PR #217, opened 2024-07-18
feat: allow Arc<String> on EncodeLabelSet

flaneur2020:
I'm writing a label struct like this:

```
#[derive(Debug, Clone, Hash, Eq, PartialEq, EncodeLabelSet)]
struct ErrorLabels {
    op: &'static str,
    scheme: &'static str,
    err: &'static str,
    root: Arc<String>,
    namespace: Arc<String>,
}
```

in this struct, the `root` and `namespace` field is nearly immutable after a client get established. i hope to use `Arc` to reduce copy Strings on updating metrics. 

but i got this error:

```
the trait bound `for<'a> &'a std::string::String: EncodeLabelValue` is not satisfied
the trait `EncodeLabelValue` is implemented for `std::string::String`
required for `Arc<std::string::String>` to implement `EncodeLabelValue`
```

it tells that we need implement &String to EncodeLabelValue.

this PR addes a implement about EncodeLabelValue for `&String`, and it can make the previous labels struct compile successfully.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support `i32`/`f32` for `Gauge` and `u32`/`f32` for `Counter`/`CounterWithExemplar`.
   See [PR 173] and [PR 216].
 
+- Supoort `Arc<String>` for `EncodeLabelValue`.
+  See [PR 217].
+
 [PR 173]: https://github.com/prometheus/client_rust/pull/173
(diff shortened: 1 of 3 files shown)


mxinden (commented):
Thanks!

Can you add a changelog entry in CHANGELOG.md?

Also, can you add a test in https://github.com/prometheus/client_rust/blob/master/derive-encode/tests/lib.rs

flaneur2020:
sure, i'll add the test and changelog in this pr!

mxinden (approved):
Thank you!

PR #218, opened 2024-07-22
Derive macro to register struct of metrics

### LINKED ISSUE #140: feat: derive register logic on struct of metrics
One oftentimes has a `struct` of metrics like:

``` rust
struct Metrics {
    a: Counter,
    b: Gauge,
}
```

Code to register the metrics in `Metrics` with a `Registry` could be generated via a derive macro on `Metrics`.
- The metric name would be taken from the field name.
- The metric help text would be taken from the Rust doc String.
- The metric unit could be provided via a macro attribute.

``` rust
#[derive(Register)]
struct Metrics {
    /// My Gauge tracking the number of xxx
    a: Counter,
    /// My Counter, tracking the event xxx
    #[prometheus-unit(Seconds)]
    b: Gauge,
}
```

Where the `Register` trait is something along the lines of:

``` rust
trait Register {
  fn register(&self, registry: &mut Registry);
}
```

Hackzzila:
Resolves #140 
cc @mxinden 

Here is a rough implementation. It does not have docs yet

Example:
```rust
#[derive(Register, Default)]
struct Metrics {
    /// This is my counter
    my_counter: Counter,
    nested: NestedMetrics,
    #[register(skip)]
    skipped: Counter,
    #[register(unit = "bytes")]
    custom_unit: Counter,
    #[register(name = "my_custom_name")]
    custom_name: Counter,
    /// This will get ignored
    #[register(help = "my custom help")]
    custom_help: Counter,
}

#[derive(Register, Default)]
struct NestedMetrics {
    /// This is my gauge
    my_gauge: Gauge,
}
```

```rust
// Only if Metrics implements default
let metrics = Metrics::register_default(&mut registry);

let metrics = Metrics::default();
metrics.register(&mut registry);
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,13 +15,14 @@ default = []
 protobuf = ["dep:prost", "dep:prost-types", "dep:prost-build"]
 
 [workspace]
-members = ["derive-encode"]
+members = ["derive-encode", "derive-register"]
 
 [dependencies]
 dtoa = "1.0"
 itoa = "1.0"
 parking_lot = "0.12"
 prometheus-client-derive-encode = { version = "0.4.1", path = "derive-encode" }
+prometheus-client-derive-register = { version = "0.1.0", path = "derive-register" }
 prost = { version = "0.12.0", optional = true }
 prost-types = { version = "0.12.0", optional = true }
 
--- a/derive-register/Cargo.toml
+++ b/derive-register/Cargo.toml
@@ -8,6 +8,9 @@ proc-macro2 = "1"
 quote = "1"
 syn = "2"
 
+[dev-dependencies]
+prometheus-client = { path = "../" }
+
 [lib]
 proc-macro = true
 
--- a/derive-register/Cargo.toml
+++ b/derive-register/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+darling = "0.20.10"
 proc-macro2 = "1"
 quote = "1"
 syn = "2"
--- a/derive-register/Cargo.toml
+++ b/derive-register/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "prometheus-client-derive-register"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+proc-macro2 = "1"
+quote = "1"
+syn = "2"
+
+[lib]
+proc-macro = true
+
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -389,3 +389,33 @@ pub trait Metric: crate::encoding::EncodeMetric + Send + Sync + std::fmt::Debug 
 
 impl<T> Metric for T where T: crate::encoding::EncodeMetric + Send + Sync + std::fmt::Debug + 'static
 {}
+
+pub trait Register {
+    fn register(self, registry: &mut Registry);
+}
+
+pub trait RegisterField {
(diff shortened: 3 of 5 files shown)


src/registry.rs
```diff
@@ -389,3 +391,51 @@ pub trait Metric: crate::encoding::EncodeMetric + Send + Sync + std::fmt::Debug
 
 impl<T> Metric for T where T: crate::encoding::EncodeMetric + Send + Sync + std::fmt::Debug + 'static
 {}
+
+pub trait Register {
```
mxinden, line 395:
```suggestion
pub trait Registrant {
```

I guess something to be registered would be a registrant. What do you think? Is that more or less intuitive? I am not a native speaker.

Cargo.toml
```diff
@@ -15,13 +15,14 @@ default = []
 protobuf = ["dep:prost", "dep:prost-types", "dep:prost-build"]
 
 [workspace]
-members = ["derive-encode"]
+members = ["derive-encode", "derive-register"]
```
mxinden, line 18: I guess we could as well do both encode and register in one derive crate? I am undecided. What is your rational for two crates?

src/registry.rs
```diff
@@ -389,3 +391,51 @@ pub trait Metric: crate::encoding::EncodeMetric + Send + Sync + std::fmt::Debug
...
+impl<T> RegisterDefault for T
+where
+    T: Register + Default,
+{
+    fn register_default(registry: &mut Registry) -> Self {
+        let this = Self::default();
+        this.register(registry);
+        this
+    }
+}
```
mxinden, line 412: Neat!

derive-register/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "prometheus-client-derive-register"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+darling = "0.20.10"
```
mxinden, line 7: I would like to keep `prometheus-client`'s dependency tree as lean as possible.

For users:

- Better compile time.
- Less crates to audit.

For maintainers:
- Less maintenance work (depending on how much we need to re-invent the wheel.)

It is not only `darling` as a dependency, but also the dependencies of `darling`.

How much work do you think would it be to implement this feature without `darling`?

mxinden (commented):
Solid work. Thank you!

I left a couple of questions.

Once we covered those, there is a bit of polishing (mostly docs).

Overall this looks good. I expect this to safe a lot of folks a lot of typing.

Cargo.toml
Hackzzila: I just went off of the existing naming. I would be fine with one crate, but then I think it should be named just `prometheus-client-derive`. Also, should this be behind a feature so people don't have to compile syn if they don't need to?

derive-register/Cargo.toml
Hackzzila: Probably only an hour or two? It doesn't seem too difficult

Cargo.toml
mxinden: - I am in favor of one crate only.
- The rename to `prometheus-client-derive` sounds good.
- Moving `prometheus-client-derive` behind a feature sounds good to me.
- I suggest making it a default feature.

derive-register/Cargo.toml
mxinden: Great. Would you mind removing `darling` from this pull request?

PR #220, opened 2024-08-01
metrics: Allow missing Gauge<u64, AtomicU64>

apelisse:


mxinden:
Thanks for the proposal.

I don't think `u64` is a valid `Gauge` value. See https://github.com/prometheus/client_rust/pull/189#issuecomment-1968792768.

apelisse:
Ah thanks!

PR #226, opened 2024-09-21
Implement Atomic<u64> for AtomicU64 for gauges

bobrik:
Between forcing end users to do endless `as i64` for things that are clearly `u64` and having one error case for rarely used protobuf when a gauge is set to `u64::MAX`, the latter seems like the right choice.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -568,6 +568,19 @@ impl EncodeGaugeValue for i64 {
     }
 }
 
+impl EncodeGaugeValue for u64 {
+    fn encode(&self, encoder: &mut GaugeValueEncoder) -> Result<(), std::fmt::Error> {
+        // Between forcing end users to do endless as i64 for things that are
+        // clearly u64 and having one error case for rarely used protobuf when
+        // a gauge is set to u64::MAX, the latter seems like the right choice.
+        if *self == u64::MAX {
+            return Err(std::fmt::Error);
+        }
+
+        encoder.encode_i64(*self as i64)
+    }
+}
+
 impl EncodeGaugeValue for f64 {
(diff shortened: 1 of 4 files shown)


src/encoding.rs
```diff
@@ -568,6 +568,16 @@ impl EncodeGaugeValue for i64 {
     }
 }
 
+impl EncodeGaugeValue for u64 {
+    fn encode(&self, encoder: &mut GaugeValueEncoder) -> Result<(), std::fmt::Error> {
+        if *self == u64::MAX {
+            return Err(std::fmt::Error);
+        }
```
mxinden, line 575: Can you add a comment here, summarizing your PR description. Makes it is easier for me in the future without going through the git history.

src/encoding/protobuf.rs
```diff
@@ -600,6 +601,42 @@ mod tests {
...
+            openmetrics_data_model::metric_point::Value::GaugeValue(value) => {
+                let expected = openmetrics_data_model::gauge_value::Value::IntValue(12345);
+                assert_eq!(Some(expected), value.value);
+            }
+            _ => panic!("wrong value type"),
+        }
+    }
+
+    #[test]
+    fn encode_gauge_u64_max() {
```
mxinden, line 631: Again, please add a small comment why this should fail.

mxinden (commented):
Thinking about this some more, I think you are right, i.e. I think this is the right tradeoff.

Thank you for being persistent. Thanks for the thorough pull request.

Two minor comments, but otherwise this is good to go. 

Also please add a changelog entry.

bobrik:
Updated. Please feel free adjust any text as you see fit.

mxinden (approved):
Thanks. @bobrik let me know if you need a release.

PR #233, opened 2024-10-11
feat: add range based exponential buckets in histogram

kohlisid:
The current implementation does not have a range based exponential bucket function. 

This can be useful in scenarios where the end user doesn't want to perform manual calculations for extracting the required buckets, especially when multiple different ranges are to be used for different histograms.

Similar implementation is provided in the golang client as well.
https://github.com/prometheus/client_golang/blob/5d584e2717ef525673736d72cd1d12e304f243d7/prometheus/histogram.go#L125

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ log = "0.4.22"
 async-std = { version = "1", features = ["attributes"] }
 axum = "0.7"
 criterion = "0.5"
+futures = "0.3"
 http-types = "2"
 pyo3 = "0.22"
 quickcheck = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ parking_lot = "0.12"
 prometheus-client-derive-encode = { version = "0.4.1", path = "derive-encode" }
 prost = { version = "0.12.0", optional = true }
 prost-types = { version = "0.12.0", optional = true }
+log = "0.4.22"
 
 [dev-dependencies]
 async-std = { version = "1", features = ["attributes"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,6 @@ parking_lot = "0.12"
 prometheus-client-derive-encode = { version = "0.4.1", path = "derive-encode" }
 prost = { version = "0.12.0", optional = true }
 prost-types = { version = "0.12.0", optional = true }
-log = "0.4.22"
 
 [dev-dependencies]
 async-std = { version = "1", features = ["attributes"] }
--- a/src/metrics/histogram.rs
+++ b/src/metrics/histogram.rs
@@ -5,6 +5,7 @@
 use crate::encoding::{EncodeMetric, MetricEncoder, NoLabelSet};
 
 use super::{MetricType, TypedMetric};
+use log::warn;
 use parking_lot::{MappedRwLockReadGuard, RwLock, RwLockReadGuard};
 use std::iter::{self, once};
 use std::sync::Arc;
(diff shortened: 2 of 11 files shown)


kohlisid:
@mxinden please check!
Thank you

src/metrics/histogram.rs
```diff
@@ -122,6 +122,27 @@ pub fn exponential_buckets(start: f64, factor: f64, length: u16) -> impl Iterato
...
 }
 
+/// Exponential bucket distribution within a range
+/// /// Creates `length` buckets, where the lowest bucket is `min` and the highest bucket is `max`.
+/// /// The final +Inf bucket is not counted and not included in the returned iterator.
+/// /// The function panics if `length` is 0 or negative, or if `min` is 0 or negative.
+fn exponential_buckets_range(min: f64, max: f64, length: u16) -> impl Iterator<Item = f64> {
+    if length < 1 {
+        panic!("ExponentialBucketsRange length needs a positive length");
+    }
```
mxinden, line 132: You could use a `NonZeroU16` instead of a `u16`. That said, that would be inconsistent with `exponential_buckets`. Thoughts?

src/metrics/histogram.rs
```diff
@@ -122,6 +122,27 @@ pub fn exponential_buckets(start: f64, factor: f64, length: u16) -> impl Iterato
         .take(length.into())
 }
 
+/// Exponential bucket distribution within a range
+/// /// Creates `length` buckets, where the lowest bucket is `min` and the highest bucket is `max`.
+/// /// The final +Inf bucket is not counted and not included in the returned iterator.
+/// /// The function panics if `length` is 0 or negative, or if `min` is 0 or negative.
```
mxinden, line 128:
```suggestion
///
/// Creates `length` buckets, where the lowest bucket is `min` and the highest bucket is `max`.
///
/// The function panics if `length` is 0 or negative, or if `min` is 0 or negative.
```

I don't think we should mention INF here, given that we don't do it anywhere else.

Also we don't do double `///` anywhere else.

src/metrics/histogram.rs
```diff
@@ -122,6 +122,27 @@ pub fn exponential_buckets(start: f64, factor: f64, length: u16) -> impl Iterato
...
+fn exponential_buckets_range(min: f64, max: f64, length: u16) -> impl Iterator<Item = f64> {
+    if length < 1 {
+        panic!("ExponentialBucketsRange length needs a positive length");
+    }
+    if min <= 0.0 {
+        panic!("ExponentialBucketsRange min needs to be greater than 0");
+    }
```
mxinden, line 135: We don't panic anywhere else in the crate. Reason being, that given that instrumentation is auxiliary only, folks likely don't want a panic. Would an empty Iterator work too?

src/registry.rs
```diff
@@ -234,7 +234,7 @@ impl Registry {
     /// let subsystem_b_counter_1: Counter = Counter::default();
     /// let subsystem_b_counter_2: Counter = Counter::default();
     ///
-    /// let subsystem_a_registry = registry.sub_registry_with_prefix("subsystem_b");
+    /// let subsystem_b_registry = registry.sub_registry_with_prefix("subsystem_b");
```
mxinden, line 237: Good catch. Nit pick, would you mind proposing this in a separate pull request?

mxinden (commented):
Neat. A couple of comments:

src/registry.rs
kohlisid: Sure! I can do that

src/metrics/histogram.rs
kohlisid: Sure

src/metrics/histogram.rs
kohlisid: I guess better to keep as u16 to keep consistent with `exponential_buckets`.
We can keep the additional check in the code for correctness.
What do you think?

src/metrics/histogram.rs
kohlisid: Sure! I have removed panics, but instead of an empty iterator I have added defaults with a warning. 
Empty iterator might have have side-effects in histogram functionality. 
Wdyt?

kohlisid:
@mxinden Thanks for the review! Have addressed them with some small follow ups, please take a look whenever you get a chance.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ parking_lot = "0.12"
 prometheus-client-derive-encode = { version = "0.4.1", path = "derive-encode" }
 prost = { version = "0.12.0", optional = true }
 prost-types = { version = "0.12.0", optional = true }
+log = "0.4.22"
```
mxinden, line 27: I try to keep the dependency footprint of this crate as small as possible. Adding `log` just for this use-case is not an option. Mind reverting it?

mxinden (commented):
Sorry for the delay here. 

Unfortunately adding a log statement is not an option.

What is the issue with returning an empty iterator?

Alternatively we can change the function to return a `Result`.

kohlisid:
Hey @mxinden 
Thanks for getting back on this
I can remove the log here for sure!
Returning the result might be cleaner but then it would diverge from the fixed buckets functions present currently, if you're okay with that we can go that route 

Otherwise I'm good with having the empty iterator as return

Let me know what works best and I can make those changes quickly and we can get this merged :)

kohlisid:
@mxinden Based on your latest addition to the change log file, could you please tell me what would be the conflict resolution here intended.

mxinden:
@kohlisid I resolved the merge conflicts. Sorry for the trouble.

Please remove the `log` dependency. You can keep the log lines as code comments. I am in favor of using an empty iterator.

kohlisid:
@mxinden Updated! Please check
Thanks

mxinden (approved):

PR #234, opened 2024-10-28
feat(metrics/family): add `get` method to Family

ethercflow:
Address comments in https://github.com/prometheus/client_rust/pull/230#issuecomment-2440804265

### DIFF
--- a/src/metrics/family.rs
+++ b/src/metrics/family.rs
@@ -226,9 +226,7 @@ impl<S: Clone + std::hash::Hash + Eq, M, C: MetricConstructor<M>> Family<S, M, C
     /// family.get_or_create(&vec![("method".to_owned(), "GET".to_owned())]).inc();
     /// ```
     pub fn get_or_create(&self, label_set: &S) -> MappedRwLockReadGuard<M> {
-        if let Ok(metric) =
-            RwLockReadGuard::try_map(self.metrics.read(), |metrics| metrics.get(label_set))
-        {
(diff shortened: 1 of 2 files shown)


ethercflow:
@mxinden PTAL, thanks!

cratelyn:
seconding this addition, i would also suppport this change, and would like have the ability to avoid creating new metrics in a family when looking up the series for a particular label set. i'll cross-reference #244, which also introduces a new accessor to `Family<S, M, C>`.

src/metrics/family.rs
```diff
@@ -452,4 +469,47 @@ mod tests {
                 .get()
         );
     }
+
+    #[test]
+    fn test_get() {
+        let family = Family::<Vec<(String, String)>, Counter>::default();
+
+        // Test getting a non-existent metric
```
mxinden, line 477:
```suggestion
        // Test getting a non-existent metric.
```

Nit, convention in the codebase, please also apply below.

mxinden (commented):
Thank you for the thorough patch.

What do you think of replacing the code below in `Family::get_or_create` with a call to `Family::get`?

```
        if let Ok(metric) =
            RwLockReadGuard::try_map(self.metrics.read(), |metrics| metrics.get(label_set))
        {
            return metric;
        }
```

Also, please add a changelog entry.

ethercflow:
@mxinden Addressed, PTAL, thanks!

mxinden (approved):
Thank you!

PR #238, opened 2024-11-10
refactor(registry): improve constructors of Registry

koushiro:
The current APIs of `Registry` does not allow me to change the `prefix` and `labels` of a `Registry` after it's constructed.

**Breaking Changes**:

- change `fn with_prefix(prefix: ...) -> Self` => `fn with_prefix(mut self, prefix: ...) -> Self`
- change `fn with_labels(labels: ...) -> Self` => `fn with_labels(mut self, labels: ...) -> Self`
- remove `fn with_prefix_and_labels(prefix: ..., labels: ...) -> Self`

### DIFF
--- a/examples/hyper.rs
+++ b/examples/hyper.rs
@@ -24,7 +24,7 @@ use tokio::{
 async fn main() {
     let request_counter: Counter<u64> = Default::default();
 
-    let mut registry = <Registry>::with_prefix("tokio_hyper_example");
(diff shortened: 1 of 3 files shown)


koushiro:
BTW, I think the `sub_registry` part is not very useful, and it might be better to remove these `sub_registry_*` APIs  to keep the `Registry` APIs simple (as long as the `prefix` and `labels` field of the registry are exposed, downstream users can implement these functionalities themselves)

PR #237, opened 2024-11-10
feat(encoding): impl `EncodeLabelValue` for `bool`

koushiro:
- impl `EncodeLabelValue` for `bool`
- adjust order of some impls (just for better consistency and readability)

PR #240, opened 2024-11-18
feat(encoding)!: 🧁 encoding traits do not consume encoders

### LINKED ISSUE #135: Allow flattening of a struct through derive(EncodeLabelSet) at any position
Currently when deriving `EncodeLabelSet` and flattening a struct the flattened struct must appear last, and there must be only one flattened struct, from the test:

```rust
#[test]
fn flatten() {
    #[derive(EncodeLabelSet, Hash, Clone, Eq, PartialEq, Debug)]
    struct CommonLabels {
        a: u64,
        b: u64,
    }
    #[derive(EncodeLabelSet, Hash, Clone, Eq, PartialEq, Debug)]
    struct Labels {
        unique: u64,
        #[prometheus(flatten)]
        common: CommonLabels,
    }

    // …
```

If you place `common` before `unique` in `struct Labels` like this:

```rust
#[derive(EncodeLabelSet, Hash, Clone, Eq, PartialEq, Debug)]
struct Labels {
    #[prometheus(flatten)]
    common: CommonLabels,
    unique: u64,
}
```

Compilation fails:

```
error[E0382]: borrow of moved value: `encoder`
   --> derive-encode/tests/lib.rs:147:14
    |
147 |     #[derive(EncodeLabelSet, Hash, Clone, Eq, PartialEq, Debug)]
    |              ^^^^^^^^^^^^^^
    |              |
    |              value borrowed here after move
    |              move occurs because `encoder` has type `LabelSetEncoder<'_>`, which does not implement the `Copy` trait
    |
```

This occurs because ownership of `encoder` is [given to the flattened struct](https://github.com/prometheus/client_rust/blob/85033bf06ae0a1ba37f5b925233282276a17a2b4/derive-encode/src/lib.rs#L40)

I think this would be a breaking change to fix, but fixing it would allow fields to appear in any order, or allow flattening of multiple structs without nesting them all one inside the other in tail position.

cratelyn:
fixes #135.

this branch updates parameter types for different encoding traits: `EncodeLabel`, `EncodeLabelSet`, `EncodeMetric`, `Collector`, and `EncodeExemplarValue`. rather than taking ownership of their respective encoder parameters, these now accept a `&mut` mutable reference.

this allows for consumers of this API to e.g. compose label sets, loosens restrictions on field order in `#[derive(EncodeLabelSet)]` structs, and provides a consistent signature type by following the pattern already set by `EncodeLabelKey`, `EncodeLabelValue`, `EncodeGaugeValue`, and other existing encoding traits.

### DIFF
--- a/derive-encode/src/lib.rs
+++ b/derive-encode/src/lib.rs
@@ -72,7 +72,7 @@ pub fn derive_encode_label_set(input: TokenStream) -> TokenStream {
 
     let gen = quote! {
         impl prometheus_client::encoding::EncodeLabelSet for #name {
-            fn encode(&self, mut encoder: prometheus_client::encoding::LabelSetEncoder) -> std::result::Result<(), std::fmt::Error> {
+            fn encode(&self, encoder: &mut prometheus_client::encoding::LabelSetEncoder) -> std::result::Result<(), std::fmt::Error> {
                 use prometheus_client::encoding::EncodeLabel;
                 use prometheus_client::encoding::EncodeLabelKey;
                 use prometheus_client::encoding::EncodeLabelValue;
--- a/examples/custom-metric.rs
+++ b/examples/custom-metric.rs
@@ -11,7 +11,7 @@ use prometheus_client::registry::Registry;
 struct MyCustomMetric {}
 
 impl EncodeMetric for MyCustomMetric {
-    fn encode(&self, mut encoder: MetricEncoder) -> Result<(), std::fmt::Error> {
+    fn encode(&self, encoder: &mut MetricEncoder) -> Result<(), std::fmt::Error> {
         // This method is called on each Prometheus server scrape. Allowing you
         // to execute whatever logic is needed to generate and encode your
         // custom metric.
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -22,20 +22,20 @@ use crate::encoding::DescriptorEncoder;
 /// struct MyCollector {}
 ///
 /// impl Collector for MyCollector {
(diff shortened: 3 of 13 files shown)


cratelyn:
i've refrained from bumping the `Cargo.toml` version in the name of offering a minimal patch, but i'm happy to bump that and add an entry to `CHANGELOG.md` if that's helpful. :slightly_smiling_face:

cratelyn:
i've also updated the benches, examples, and doctests so this should be ready for review now. :hammer:

cratelyn:
this work conflicts with the [cosmetic changes](https://github.com/prometheus/client_rust/pull/237#discussion_r1847437725) that were made in #237. i'm happy to rebase this branch and address those conflicts, but i'd like to hold off on doing so until i have a signal that we'd be disposed to merge these changes.

please ping me if, or when, that work would be desired.

mxinden:
Thinking about this some more, I think the choice of taking ownership was deliberate when I designed these builder patterns.

Can you come up with an example that is not supported by the status quo?

olix0r:
Yes, It's not possible to pair two arbitrary labelsets, for example. We end up with this monstrosity that we have to implement on all of our EncodeLabelSet impls

https://github.com/linkerd/linkerd2-proxy/blob/fed1e89a01164977661dbc95f1084f5b40b55ec7/linkerd/metrics/src/lib.rs#L47-L49

Which is needed to implement something like:
https://github.com/linkerd/linkerd2-proxy/blob/fed1e89a01164977661dbc95f1084f5b40b55ec7/linkerd/app/outbound/src/opaq/concrete.rs#L85-L93

cratelyn:
succeeded by #257.

PR #242, opened 2024-11-18
feat(metrics/histogram): 🍪 `count()` and `sum()` accessors

### LINKED ISSUE #241: `Histogram` does not provide accessors
#### :speech_balloon: background

`Counter` and `Gauge` provide a `get()` method, which are helpful for various use cases. currently, `Histogram` does not provide any equivalent way for callers to observe its count, sum, or buckets.

for my particular use case, i am interested in test coverage that confirms that Prometheus metrics are properly installed and measuring behavior in my application. because there is not any way to inspect histograms however, i am left in an unfortunate position where i can only write test coverage for my application if i use counters and gauges.

so, i'd like to propose the addition of a few new methods to `Histogram`.

#### :balance_scale: proposal (part 1)

currently the histogram structure is defined as such:

```rust
pub struct Histogram {
    inner: Arc<RwLock<Inner>>,
}


#[derive(Debug)]
pub(crate) struct Inner {
    sum: f64,
    count: u64,
    buckets: Vec<(f64, u64)>,
}
```

first, i'd like to propose the following, which seem least controversial and most straightforward:

```rust
impl Histogram {
    /// Returns the current sum of all observations.
    pub fn sum(&self) -> f64 {
        self.inner.read().sum
    }

    /// Returns the current number of observations.
    pub fn count(&self) -> u64 {
        self.inner.read().count
    }
}
```

these two methods would address most of my particular concerns with respect to introspection in tests, and would help bring `Histogram` in line with the interfaces provided by `Gauge` and `Counter` metrics.

#### :balance_scale: proposal (part 2)

it would also be very helpful if there was a way to access the buckets.

most simply, we could provide a way to acquire an iterator over the bucket tuples. i'd imagine this would look approximately like:

```rust
impl Histogram {
    /// Returns a view of the histogram's buckets.
    pub fn buckets(&self) -> MappedRwLockReadGuard<[(f64, u64)]> {
        let inner = self.inner.read();
        let buckets = RwLockReadGuard::map(inner, |inner| inner.buckets.as_slice());
        buckets
    }
}
```

we could provide a way to get the bucket for a given observed value. the nice part about this is that a caller wouldn't need to know how a histogram has been configured, and spares them the trouble/boilerplate of iterating through all of the buckets to find the proper bucket.

```rust
impl Histogram {
    /// Returns the bucket given an observed value.
    pub fn lookup(&self, v: f64) -> Option<(f64, u64)> {
        let inner = self.inner.read();
        inner
            .buckets
            .iter()
            .find(|(upper_bound, _value)| upper_bound >= &v)
            .copied()
    }
}
```

note that this isn't mutually exclusive with the `buckets()` accessor, we could provide either/both of these.

that said, i'm less attached to these bucket interfaces. i'll also note that the `MappedRwLockReadGuard` would be subject to the same deadlock concerns mentioned in #231, but this signature seems like it follows the spirit of other existing interfaces in this library.

---

i am **especially** interested in adding `sum()` and `count()`, though i am also happy to help with adding the other methods if these are also welcome.

cratelyn:
fixes #241.

this commit introduces two new public methods to `Histogram`; `sum()` and `count()` return the sum of all observations and the number of observations made, respectively.


cratelyn:
:wave: hello! this pull request in particular would be very helpful for our usage of `Histogram`s. i would love if you could take a look at this soon, i believe it's a small and non-controversial addition to the `Histogram` interface.

mxinden:
I am hesitant exposing the internals of `Histogram` just for the use-case of testing. Unless we come up with better use-cases, I don't think we should merge here.

olix0r:
@mxinden I'm sympathetic to maintaining a minimal public API, but it's pretty important to us that we can write tests that ensure that our metrics are properly instrumented. For example:

https://github.com/linkerd/linkerd2-proxy/blob/ea5136268d65aa6a570cb612150875f923067042/linkerd/app/outbound/src/http/logical/policy/route/backend/metrics/tests.rs#L224-L237

Would you be amenable to feature-gating these public helpers behind a `test-util` feature? This is a pretty common approach we've seen in other crates.

Or are there alternative approaches you recommend? Would it be better to expose something a module akin to [`github.com/prometheus/client_golang/prometheus/testutil`](https://pkg.go.dev/github.com/prometheus/client_golang@v1.20.5/prometheus/testutil) to make assertions about a registry?

antoniovicente:
> @mxinden I'm sympathetic to maintaining a minimal public API, but it's pretty important to us that we can write tests that ensure that our metrics are properly instrumented. For example:
> 
> https://github.com/linkerd/linkerd2-proxy/blob/ea5136268d65aa6a570cb612150875f923067042/linkerd/app/outbound/src/http/logical/policy/route/backend/metrics/tests.rs#L224-L237
> 
> Would you be amenable to feature-gating these public helpers behind a `test-util` feature? This is a pretty common approach we've seen in other crates.
> 
> Or are there alternative approaches you recommend? Would it be better to expose a module akin to [`github.com/prometheus/client_golang/prometheus/testutil`](https://pkg.go.dev/github.com/prometheus/client_golang@v1.20.5/prometheus/testutil) to make assertions about a registry?

I'm running into this same class of problems: Histogram does not provide an easy way to check the contents in a test.  I have no way to test my code that updates histograms based on the result of some computation.

The way to extract data from histograms doesn't need to be performant, but it does need to exist.  It could be something that requires serializing to a string and back.

cratelyn:
i've opened #283 to introduce a `test-util` feature. i can rebase this, and other proposals that received similar feeback, upon that if/when it has landed.

cratelyn:
i updated this branch, and ci is now passing.

cratelyn:
let me know if there is anything i can do to help push this forward, i'm interested in unblocking the possibility for testing metrics for the reasons outlined [above](https://github.com/prometheus/client_rust/pull/242#issuecomment-2542031813). thanks for your time.

mxinden (approved):
Change looks good to me. Thank you.

@cratelyn I assume you still want me to merge here?

PR #243, opened 2024-11-18
refactor(metrics/histogram): 🎂 constructor accepts `IntoIterator`

cratelyn:
this is a very small, **non-breaking**, alteration to the signature of `Histogram`'s constructor.

rather than accepting an `impl Iterator`, this commit changes the parameter of `Histogram::new()` to be an `impl IntoIterator` instead. this does not affect the existing contract because of the blanket `impl<I: Iterator> IntoIterator for I` implementation provided by the standard library.

by accepting `IntoIterator` however, callers providing a collection such as a `[f64; 5]` array or a `Vec<f64>` vector do not need to invoke `into_iter()` themselves at the call-site.

```rust
// now, constructing a histogram needn't involve `into_iter()`...
use prometheus_client::metrics::histogram::Histogram;
let histogram = Histogram::new([10.0, 100.0, 1_000.0]);
```

this leans on the same sugar used by `for {}` loops, see the relevant section of the `std::iter` documentation here:
<https://doc.rust-lang.org/stable/std/iter/index.html#for-loops-and-intoiterator>

no changes are needed within `Histogram::new()` because we already call `into_iter()` on the provider iterator when appending `f64::MAX` and collecting the buckets into a `Vec<_>`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.24.0] - unreleased
+## [0.23.1] - unreleased
+
+### Changed
 
 - `Histogram::new` now accepts an `IntoIterator` argument, rather than an `Iterator`.
   See [PR 243].
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.24.0] - unreleased
+
+- `Histogram::new` now accepts an `IntoIterator` argument, rather than an `Iterator`.
+  See [PR 243].
+
+[PR 243]: https://github.com/prometheus/client_rust/pull/243
+
 ## [0.23.0]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


mxinden (commented):
This looks good to me. Neat!

@cratelyn would you mind adding a changelog entry?

cratelyn:
sure thing! i've rebased this on the latest `master` branch, and have added a changelog entry for this pull request. how does 1e216f1 look?

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.24.0] - unreleased
```
```suggestion
## [0.23.1]
```

Not a breaking change.

Also, if you bump the version in `Cargo.toml` to `v0.24.1` I am happy to cut a patch release right away.

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
...
+
+- `Histogram::new` now accepts an `IntoIterator` argument, rather than an `Iterator`.
```
mxinden, line 9:
```suggestion
### Added

- `Histogram::new` now accepts an `IntoIterator` argument, rather than an `Iterator`.
```

See keepachangelog format.

CHANGELOG.md
cratelyn: ce22f2d changes the version in the changelog. it also bumps the version in the manifest to `0.23.1` (_not `0.24.1`, to be clear_).

CHANGELOG.md
cratelyn: sure. ce22f2d adds a header. note that it adds a `Changed` header, not an `Added` header, because the `Histogram` constructor was not added here.

mxinden (approved):
Thanks!

mxinden:
Tagged and published.

PR #244, opened 2024-11-18
feat(metrics/family): 🍫 add `Family::get_or_create_owned()`

### LINKED ISSUE #231: Deadlock when using `Family::get_or_create`
It's fairly common practice for us to create a set of metrics from a metric family given a set of labels, and store them for later use. However, this can cause a deadlock when `Family::get_or_create` is called multiple times on the same instance within the same statement:

```rust
struct Metrics {
  inbound: Counter,
  outbound: Counter,
}

let family = Family::<Vec<(String, String)>, Counter>::default();

Metrics {
  inbound: family.get_or_create(&vec!["dir", "inbound"]).clone(),
  outbound: family.get_or_create(&vec!["dir", "outbound"]).clone(), // deadlocks
}
```

This is due to the `RwLockReadGuard` returned by each call to `Family::get_or_create` being temporaries that are not dropped until the end of the statement.

The user-facing solution is to create separate bindings for each metric:

```rust
// ...

let inbound - family.get_or_create(&vec!["dir", "inbound"]).clone();
let outbound = family.get_or_create(&vec!["dir", "outbound"]).clone(); // no deadlock
Metrics { inbound, outbound }
```

However, this is a footgun that can be pretty tricky to debug. Is there an alternative API here that would work, perhaps just returning the metric directly instead of a mutex guard?

cratelyn:
fixes #231.

this introduces a new method to `Family<S, M, C>` to help alleviate the risk of deadlocks when accessing multiple series within a given metrics family.

this returns a plain `M`, rather than the `MappedRwLockReadGuard<M>` RAII guard returned by `get_or_create()`.

a test case is introduced in this commit to demonstrate that structures accessing multiple series within a single expression will not accidentally create a deadlock.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    method avoids the risk of runtime deadlocks at the expense of creating an
    owned type. See [PR 244].
  
- [PR 244]: https://github.com/prometheus/client_rust/pull/244
+[PR 244]: https://github.com/prometheus/client_rust/pull/244
+[PR 257]: https://github.com/prometheus/client_rust/pull/257
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `EncodeLabelSet` is now implemented for tuples `(A: EncodeLabelSet, B: EncodeLabelSet)`.
    See [PR 257].
 
-- `Family::get_or_create_clone` can access a metric in a labeled family. This
+- `Family::get_or_create_owned` can access a metric in a labeled family. This
    method avoids the risk of runtime deadlocks at the expense of creating an
    owned type. See [PR 244].
+ 
+ [PR 244]: https://github.com/prometheus/client_rust/pull/244
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `EncodeLabelSet` is now implemented for tuples `(A: EncodeLabelSet, B: EncodeLabelSet)`.
+   See [PR 257].
+
+- `Family::get_or_create_clone` can access a metric in a labeled family. This
+   method avoids the risk of runtime deadlocks at the expense of creating an
+   owned type. See [PR 244].
 
 ### Changed
 
--- a/src/metrics/family.rs
+++ b/src/metrics/family.rs
@@ -261,7 +261,7 @@ impl<S: Clone + std::hash::Hash + Eq, M, C: MetricConstructor<M>> Family<S, M, C
     /// ```
     ///
     /// NB: This method can cause deadlocks if multiple metrics within this family are read at
-    /// once. Use [`Family::get_or_create()`] if you would like to avoid this by cloning the
+    /// once. Use [`Family::get_or_create_clone()`] if you would like to avoid this by cloning the
     /// metric `M`.
     pub fn get_or_create(&self, label_set: &S) -> MappedRwLockReadGuard<M> {
         if let Some(metric) = self.get(label_set) {
--- a/src/metrics/family.rs
+++ b/src/metrics/family.rs
@@ -207,6 +207,40 @@ impl<S: Clone + std::hash::Hash + Eq, M, C> Family<S, M, C> {
     }
 }
 
+impl<S: Clone + std::hash::Hash + Eq, M: Clone, C: MetricConstructor<M>> Family<S, M, C>
+where
+    S: Clone + std::hash::Hash + Eq,
+    M: Clone,
+    C: MetricConstructor<M>,
+{
+    /// Access a metric with the given label set, creating it if one does not yet exist.
+    ///
+    /// ```
+    /// # use prometheus_client::metrics::counter::{Atomic, Counter};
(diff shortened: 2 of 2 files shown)


cratelyn:
i've rebased this commit to address merge conflicts, and added information to the changelog. let me know if there is anything else that needs to be done. i am keen on a release that addresses the deadlocks reported in #231.

src/metrics/family.rs
```diff
@@ -207,6 +207,40 @@ impl<S: Clone + std::hash::Hash + Eq, M, C> Family<S, M, C> {
...
+    /// // Will create and return the metric with label `method="GET"` when first called.
+    /// family.get_or_create_clone(&vec![("method".to_owned(), "GET".to_owned())]).inc();
+    ///
+    /// // Will return a clone of the existing metric on all subsequent calls.
+    /// family.get_or_create_clone(&vec![("method".to_owned(), "GET".to_owned())]).inc();
+    /// ```
+    ///
+    /// Callers wishing to avoid a clone of the metric `M` can call [`Family::get_or_create()`] to
+    /// return a reference to the metric instead.
+    pub fn get_or_create_clone(&self, label_set: &S) -> M {
```
mxinden, line 233:
```suggestion
    pub fn get_or_create_owned(&self, label_set: &S) -> M {
```

Would this be more intuitive? I am undecided.

mxinden (commented):
Thank you for working on this.

src/metrics/family.rs
```diff
@@ -225,6 +259,10 @@ impl<S: Clone + std::hash::Hash + Eq, M, C: MetricConstructor<M>> Family<S, M, C
     /// // calls.
     /// family.get_or_create(&vec![("method".to_owned(), "GET".to_owned())]).inc();
     /// ```
+    ///
+    /// NB: This method can cause deadlocks if multiple metrics within this family are read at
+    /// once. Use [`Family::get_or_create()`] if you would like to avoid this by cloning the
```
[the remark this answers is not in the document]
cratelyn: ah! good catch, 5b097b0 fixes this :white_check_mark:

src/metrics/family.rs
cratelyn: i'm not strongly opinionated about the name, i went with `get_or_create_clone()` per your [previous suggestion](https://github.com/prometheus/client_rust/issues/231#issuecomment-2440130338). i'm happy with either. :slightly_smiling_face: :+1:

my ideal, personally, would be the interface described [here](https://github.com/prometheus/client_rust/issues/231#issuecomment-2415001551), because exposing read and write guards is prone to deadlocks, but that would be a breaking change at this point.

mxinden: Given that we call `to_owned`, I think the rename suggest above is more accurate. Thus I am in favor of `get_or_create_owned`. Sorry for the trouble. Mind doing the rename?

CHANGELOG.md
```diff
@@ -4,7 +4,15 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.23.1] - unreleased
+## [0.23.2]
+
+### Added
+
+- `Family::get_or_create_clone` can access a metric in a labeled family. This
+   method avoids the risk of runtime deadlocks at the expense of incrementing
+   a reference count.
```
mxinden, line 13:
```suggestion
   method avoids the risk of runtime deadlocks at the expense of creating an owned type.
```

Given that `M` is generic, calling `to_owned` on it might use some other synchronization mechanism than reference counting.

mxinden (commented):
Two more comments. Otherwise looks good to me.

Agreed, that this is not the ideal solution, but in my eyes a good compromise, given that it is not a breaking change. Long term, we can redesign this interface in a breaking release.

CHANGELOG.md
cratelyn: this wording is addressed in 4c24551.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `EncodeLabelSet` is now implemented for tuples `(A: EncodeLabelSet, B: EncodeLabelSet)`.
+   See [PR 257].
+
+- `Family::get_or_create_clone` can access a metric in a labeled family. This
```
mxinden, line 14:
```suggestion
- `Family::get_or_create_owned` can access a metric in a labeled family. This
```

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+   method avoids the risk of runtime deadlocks at the expense of creating an
+   owned type. See [PR 244].
```
mxinden, line 16:
```suggestion
   owned type. See [PR 244].
 
 [PR 244]: https://github.com/prometheus/client_rust/pull/244
```

CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `EncodeLabelSet` is now implemented for tuples `(A: EncodeLabelSet, B: EncodeLabelSet)`.
+   See [PR 257].
+
+- `Family::get_or_create_owned` can access a metric in a labeled family. This
+   method avoids the risk of runtime deadlocks at the expense of creating an
+   owned type. See [PR 244].
+ 
+ [PR 244]: https://github.com/prometheus/client_rust/pull/244
```
mxinden, line 18:
```suggestion
[PR 244]: https://github.com/prometheus/client_rust/pull/244
[PR 257]: https://github.com/prometheus/client_rust/pull/257
```

mxinden (approved):

PR #249, opened 2024-11-28
feat(metrics): add Summary implementation

zhangtianhao:
This commit introduces the `Summary` metric type for the Rust Prometheus client, addressing the requirements outlined in https://github.com/prometheus/client_rust/issues/40.

### Key Highlights:
- Implements the `Summary` type to represent summary statistics such as sum, count, and quantiles.
- The implementation does not include a specific quantile calculation algorithm. Users are expected to provide precomputed quantiles based on their chosen algorithm.
- Supports encoding summary metrics following the OpenMetrics standard.

### Details:
- `Summary` provides:
  - `reset` method to update its internal state with sum, count, and quantiles.
  - Integration with the `TypedMetric` and `EncodeMetric` traits to support encoding and registration with the Prometheus client.

- Documentation and tests have been added to ensure correct usage and behavior.
- This implementation is designed to give users flexibility in handling quantile computation.

### Notes:
- Future work could explore integrating popular quantile calculation algorithms to offer built-in support while retaining the option for custom implementations.

Signed-off-by: zth <zhangtianhao30@gmail.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,26 @@
+# Contributing
+
+## Protocol Buffers
+
+The `build.rs` script in this library depends upon the
+[Protocol Buffers compiler][protoc]. Be sure that `protoc` is installed and
+available within your `PATH`.
+
+[protoc]: https://docs.rs/prost-build/latest/prost_build/#sourcing-protoc
+
+## Python Dependencies
+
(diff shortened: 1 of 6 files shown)


mxinden:
@zhangtianhao why close here? Are you no longer interested in contributing this feature?

PR #257, opened 2025-01-21
feat(encoding)!: `EncodeLabelSet::encode()` uses reference

cratelyn:
this commit alters the signature of the `EncodeLabelSet::encode()` trait method, such that it now accepts a mutable reference to its encoder.

this is related to #135, and is a second proposal following previous work in #240.

this change permits distinct label sets to be composed together, now that the label set encoder is not consumed. a new implementation for tuples `(A, B)` is provided.

this commit includes a test case showing that a metric family can compose two label sets together, and that such a family can successfully be digested by the python client library.

`derive-encode` is altered to generate code matching this new trait signature, and has been bumped to version 0.5.0 as a result of this breaking change in the `prometheus-client` library.

### DIFF
--- a/src/encoding/text.rs
+++ b/src/encoding/text.rs
@@ -296,8 +296,9 @@ pub(crate) struct MetricEncoder<'a> {
 impl std::fmt::Debug for MetricEncoder<'_> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let mut labels = String::new();
+        let mut encoder = LabelSetEncoder::new(&mut labels).into();
         if let Some(l) = self.family_labels {
-            l.encode(LabelSetEncoder::new(&mut labels).into())?;
+            l.encode(&mut encoder)?;
         }
 
         f.debug_struct("Encoder")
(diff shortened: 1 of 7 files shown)


src/encoding/text.rs
```diff
@@ -1114,6 +1115,68 @@ mod tests {
...
+            let labels = (Color("red"), Size("large"));
+            let counter = family.get_or_create(&labels);
+            counter.inc();
+        }
+        /*{
+            XXX: this test cannot deterministically check if multiple label values work. see #125.
+            let labels = (Color("blue"), Size("small"));
+            let counter = family.get_or_create(&labels);
+            counter.inc_by(2);
+        }*/
```
mxinden, line 1166: Is any code-path of this new feature untested without this?

I guess we could always compare to both possible outcome strings. That said, I don't think it is necessary.

mxinden (commented):
Very well done @cratelyn! Thank you.

@olix0r want to give this a review as well? Does this address your issue mentioned in https://github.com/prometheus/client_rust/pull/240#issuecomment-2593653847? Are there any other (breaking) changes you would need for linkerd?

olix0r (approved):
lgtm! thanks @cratelyn @mxinden :taco::taco:

src/encoding/text.rs
cratelyn: okay, 98bca72 removes this.

mxinden (approved):
Thanks!

Issue #258, opened 2025-02-11
Encoding to std::io::Write?


RandomInsano:
I'm providing data directly to a socket. It looks like `encode()` writes to `std::fmt::Write` so it must be written to buffer then sent to the socket.

That's serviceable, but seems like a bit of a waste of memory and performance. So, two questions:

1. Is there a way to write to std::io::Write that I haven't seen? Reviewing the text encoder, it seems like no
2. Is there any interest in a pull request that would implement this? Maybe as `encode_stream()` and hide that behind the existing `encode()` and friends?

mxinden:
I am not aware of an official adapter between the two traits. That said, what is stopping you from writing your own?

``` rust
use std::io;
use std::fmt;

pub struct IoFmtWriter<W: io::Write>(pub W);

impl<W: io::Write> fmt::Write for IoFmtWriter<W> {
    fn write_str(&mut self, s: &str) -> fmt::Result {
        self.0.write_all(s.as_bytes()).map_err(|_| fmt::Error)
    }
}
```

If that is indeed an option for you, I would assume others are facing the same issue. If so, would you want to create a pull request with an example to this repository?

RandomInsano:
Yeah, that's an excellent idea! I actually moved to using Tokio so I'll include both a sync and asynchronous example. I'll add it to next-week's TODO list.

RandomInsano:
I found it quite difficult to get the await into `write_str` so instead of delay further there's just the synchronous version for now.

RandomInsano:
Example is merged. Closing

PR #261, opened 2025-02-24
Add example using std::io::Write

RandomInsano:
Help users learn how to directly write data to an object implementing `std::io::Write`.

Doing some tests with 1,000 counters:
* Directly to a String: 454KB of RAM
* Directly to a file: 305KB of RAM

Closes out #258

### DIFF
--- a/examples/io_write.rs
+++ b/examples/io_write.rs
@@ -1,0 +1,39 @@
+//! Example showing how one could write to a file or socket instead of a string.
+//! For large metrics registries this will be more memory efficient.
+
+use prometheus_client::{encoding::text::encode, metrics::counter::Counter, registry::Registry};
+use std::io::Write;
+
+fn main() {
+    let mut registry = <Registry>::with_prefix("stream");
+    let request_counter: Counter<u64> = Default::default();
+
+    registry.register(
+        "requests",
+        "How many requests the application has received",
+        request_counter.clone(),
+    );
+
+    let mut buf = String::new();
+    encode(&mut buf, &registry).unwrap();
+
+    let mut file = Vec::new();
+    let mut writer = IoWriterWrapper(&mut file);
+    encode(&mut writer, &registry).unwrap();
+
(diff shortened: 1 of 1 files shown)


RandomInsano:
For the testing, I used `heaptrack` on Linux and read the "peak heap memory consumption" value. One one test of each was done.

examples/io_write.rs
```diff
@@ -0,0 +1,40 @@
...
+    let mut file = Vec::new();
+    let mut writer = IoWriterWrapper(&mut file);
+    encode(&mut writer, &registry).unwrap();
+
+    assert!(buf.as_bytes() == file);
+}
+
+pub struct IoWriterWrapper<W>(W)
+where
+    W: Write;
```
mxinden, line 28:
```suggestion
pub struct IoWriterWrapper<W>(W);
```

Is the bound needed in the struct definition? It should only be needed in the `std::fmt::Write` implementation below.

examples/io_write.rs
```diff
@@ -0,0 +1,40 @@
+use prometheus_client::{encoding::text::encode, metrics::counter::Counter, registry::Registry};
+use std::io::Write;
+
+/// Example showing how one could write to a file or socket instead of a string.
+/// For large metrics registries this will be more memory efficient
```
mxinden, line 5: I would assume a module comment would be more appropriate. Feel free to ignore.

```suggestion
//! Example showing how one could write to a file or socket instead of a string.
//! For large metrics registries this will be more memory efficient.
```

mxinden:
@RandomInsano just to avoid any confusion, you would need to make the above changes and then push another commit or amend your existing commit. Of course, there is no rush.

RandomInsano:
Thanks for the clarification! Code is pushed now. Verified the code still compiles and the assert still passes.

RandomInsano:
Missed the DCO failure. Decided to squash and force a new commit.

mxinden (approved):
Well done. Thanks.

PR #270, opened 2025-04-19
feat(prometheus-client-derive): initial implemenation

### LINKED ISSUE #140: feat: derive register logic on struct of metrics

ADD-SP:
## Summary

This PR implements features proposed by #140, supersedes #218 (which doesn't makes new progress in the past several months).

This PR adds and publishes (to `crates.io`) a new crate `prometheus-client-derive`, and re-exports its macros in crate `prometheus-client`, so the user should use `prometheus-client` instead of `prometheus-client-derive`. This pattern is referenced by `serde`.

The new crate only has one macros `Registrant` derive the trait that allows structs to be registered in a registry.

## Futures of `derive-encode`

This PR doesn't touch it, but I think we should merge it into `prometheus-client-derive` eventually as the first one's functionalities is a subset of the second one.

## Issues

close #140, supersede #218

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,15 @@ default = []
 protobuf = ["dep:prost", "dep:prost-types", "dep:prost-build"]
 
 [workspace]
-members = ["derive-encode"]
+members = ["derive-encode", "prometheus-client-derive"]
+
+[workspace.dependencies]
+proc-macro2 = "1"
+quote = "1"
+syn = "2"
+
+# dev-dependencies
+trybuild = "1"
 
 [dependencies]
 dtoa = "1.0"
@@ -24,6 +32,7 @@ parking_lot = "0.12"
 prometheus-client-derive-encode = { version = "0.5.0", path = "derive-encode" }
 prost = { version = "0.12.0", optional = true }
 prost-types = { version = "0.12.0", optional = true }
+prometheus-client-derive = { version = "0.24.0", path = "./prometheus-client-derive" }
 
 [dev-dependencies]
 async-std = { version = "1", features = ["attributes"] }
@@ -40,6 +49,7 @@ tokio = { version = "1", features = ["rt-multi-thread", "net", "macros", "signal
 hyper = { version = "1.3.1", features = ["server", "http1"] }
 hyper-util = { version = "0.1.3", features = ["tokio"] }
 http-body-util = "0.1.1"
+prometheus-client-derive = { path = "./prometheus-client-derive" }
 
 [build-dependencies]
 prost-build = { version = "0.12.0", optional = true }
--- a/prometheus-client-derive/src/registrant/attribute.rs
+++ b/prometheus-client-derive/src/registrant/attribute.rs
@@ -1,0 +1,146 @@
+use quote::ToTokens;
+use syn::spanned::Spanned;
+use proc_macro2::Span;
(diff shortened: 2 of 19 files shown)


Cargo.toml
```diff
@@ -24,6 +32,7 @@ parking_lot = "0.12"
 prometheus-client-derive-encode = { version = "0.5.0", path = "derive-encode" }
 prost = { version = "0.12.0", optional = true }
 prost-types = { version = "0.12.0", optional = true }
+prometheus-client-derive = { version = "0.24.0", path = "./prometheus-client-derive" }
```
ADD-SP, line 35: Shall we gate this dependency using a feature flag?

prometheus-client-derive/src/registrant/attribute.rs
```diff
@@ -0,0 +1,135 @@
...
+                                meta.path.span(),
+                                "Duplicate `unit` attribute",
+                            ));
+                        }
+
+                        // unit should be lowercase
+                        let unit = syn::LitStr::new(
+                            unit.value().as_str().to_ascii_lowercase().as_str(),
+                            unit.span(),
+                        );
```
ADD-SP, line 101: There are two options:

1. Covert to lowercase silently.
2. Emit compilation error.

What do you think?

prometheus-client-derive/src/registrant/attribute.rs
```diff
@@ -0,0 +1,135 @@
...
+                        attr.unit = Some(unit);
+                    } else if meta.path.is_ident("rename") {
+                        let rename = meta.value()?.parse::<syn::LitStr>()?;
```
ADD-SP, line 104: Shall we emit compilation error if the name is not in snake_case?

prometheus-client-derive/src/registrant/attribute.rs
```diff
@@ -0,0 +1,135 @@
...
+use quote::ToTokens;
+use syn::spanned::Spanned;
+
+// do not derive debug since this needs "extra-traits"
+// feature for crate `syn`, which slows compile time
+// too much, and is not needed as this struct is not
+// public.
+#[derive(Default)]
+pub struct Attribute {
+    pub help: Option<syn::LitStr>,
```
ADD-SP, line 11: Shall we make the `help` as a required field? Lack of help message might make no sense.

koushiro:
Interesting idea, would you be interested in porting this PR to [my openmetrics library](https://github.com/koushiro/fastmetrics)?

ADD-SP:
@mxinden Could you take a look?

PR #273, opened 2025-07-22
feat: impl<T: Collector> Collector for std::sync::Arc<T>

hanabi1224:
This PR implements `Collector` for `std::sync::Arc<T>` where `T: Collector`. This is useful for registering an `Arc` pointer of the metrics tracker instead of wrapping tracker fields with `Arc`


mxinden (commented):
Change looks good to me. Thanks.

You would need to add a `git --signoff` to your commit in order for us to merge.

PR #279, opened 2025-08-12
Avoid encoding descriptor of empty family

### LINKED ISSUE #271: Empty metric result when no value is observed
We are running into an issue with the metrics output where a metric that had no observations gets output to the file _without_ the value line. 

```
# HELP example_one Some counter metric 1.
# TYPE example_one counter
# HELP example_two Some gauge metric 2.
# TYPE example_two gauge
# EOF
```

I ran the same example using the [golang client](https://github.com/prometheus/client_golang) and for labeled metrics they are hidden from the output when there are no observations. My expectation is that this would behave the same.

Note: metrics without labels get zeroed out when there are no observations. 

Here's the code I wrote to reproduce this:

```rust
#[derive(Default, Clone, Debug, Hash, PartialEq, Eq, EncodeLabelSet)]
struct Label {
    pub key: String,
}

fn main() {
    let example1 = Family::<Label, Counter>::default();
    let example2 = Family::<Label, Gauge>::default();

    registry.register(
        "example_one",
        "Some counter metric 1",
        example1.clone(),
    );
    registry.register(
        "example_two",
        "Some gauge metric 2",
        example2.clone(),
    );


    // Do not observe the metric!

    let mut buffer = String::new();

    encode(&mut buffer, &registry).unwrap();

    println!("{}", buffer);
}
``` 

Is this a bug or is it the intended behavior?

spoutn1k:
Fixes #271. The implementation should be extensible and a user choice when implementing `is_empty` on the `EncodeMetric` trait. This disables metric descriptor generation on families. The original issue does not make clear if this is an implementation choice or a bug, I can feature-gate the change if this is a choice.


mxinden (commented):
Thanks for the patch.

Needs a test and a changelog entry.

In addition, can you point to the corresponding code in the Golang client?

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
```
```suggestion
## [0.24.1]
```

mxinden:
I don't have permissions to apply the above suggestions. Mind granting me those permissions, or applying them yourself?

mxinden (approved):
Thank you!

PR #281, opened 2025-08-30
refactor(encoding): remove `as` casts

jalil-salame:
There was a bug in the `impl EncodeGaugeValue for u64` implementation due to the `as i64` cast.

The implementation only checked if `self == u64::MAX`, this is wrong since `u64::MAX` is not the only value that cannot be represented by an `i64`.

The range of values is actually `self > i64::MAX`. `u64::MAX == 2^64 - 1` whereas `i64::MAX == 2^63 - 1`, this means there are `2^63` `u64` values that are not representable by an `i64`, not just `i64::MAX`. Delegating the checks to `i64::try_from(self)` ensures the logic is right. For this reason I also switched the rest of the implementations to use `N::from` instead of `as N` since the `N::from` function is only implemented for types whose conversion is infallible, this guarantees no such issues will ever happen.

The only `as` cast left is `u64 as f64` in `EncodeExemplarValue`, this is not possible to remove since there is no `impl TryFrom<u64> for f64`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.24.1]
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


jalil-salame:
Also created https://github.com/jalil-salame/client_rust/pull/1 as a follow up (adding `EncodeGaugeValue` and `EncodeCounterValue` impls for `usize` and `isize`).

mxinden (approved):

PR #291, opened 2025-12-30
feat(example): add an axum example with a global registry and a global metric

BentEngbers:
I am currently migrating a large project from [prometheus](https://github.com/tikv/rust-prometheus) to this crate.

Given the high volume of metrics and the variety of tasks in our project, we have opted for a **global registry and global metrics**. I have modified the [examples/axum.rs](https://github.com/prometheus/client_rust/blob/master/examples/axum.rs) example to demonstrate this pattern and provide a reference for other users.

1. Are there any improvements or best practices you would suggest for this example?
2. Are there potential changes to the library that could better support this specific use case?

### DIFF
--- a/examples/axum_with_globals.rs
+++ b/examples/axum_with_globals.rs
@@ -19,7 +19,7 @@ use tokio::time;
 
 static REGISTRY: LazyLock<Mutex<Registry>> = LazyLock::new(|| Mutex::new(Registry::default()));
 
-fn register_metric_to_global_registry<MetricType: Metric + Clone + Default>(
+pub fn register_metric_to_global_registry<MetricType: Metric + Clone + Default>(
     name: &str,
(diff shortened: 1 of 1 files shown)


cormacrelf:
> Are there any improvements or best practices you would suggest for this example?

Well you should note that if you have a LazyLock<Counter>, and you never increment it, you will never see `my_counter 0` in the output due to the lazy init never having needed to run.

I guess if this is a problem you could work around it by using https://docs.rs/inventory to build a global list of `&'static` reference to metrics that can be registered along with name/description/unit, and iterate that at server startup. Looks very difficult to get the `.register` function to work, but there is an EncodeMetric trait which could be implemented on `LazyLock<T> where T: EncodeMetric` or perhaps `&'static LazyLock<T>`, something like that.

PR #295, opened 2026-03-30
protobuf: allow compiling without protoc

howardjohn:
Our goal is to have our builds compile without external dependencies.
`protox` helps achieve this.


mxinden:
@howardjohn do I understand correctly, that you are actually using the Prometheus protobuf format? I wasn't aware of any users thus far, and was thus actually considering removing protobuf support all together from `client_rust`.

howardjohn:
@mxinden no, I plan to adopt it though with the ultimate goal of getting native histograms.

howardjohn:
replaced by https://github.com/prometheus/client_rust/pull/298

PR #294, opened 2026-03-27
dependencies: update prost to 0.14

howardjohn:

PR #298, opened 2026-04-01
protobuf: move from OpenMetrics to Prometheus proto

### LINKED ISSUE #296: protobuf: replace/augment OpenMetrics protobuf with Prometheus format
Currently, this library supports OpenMetrics proto encoding. However, it does not support Prometheus protobuf encoding.

This seems reasonable on the surface as it mirrors the text encoding, but (IIUC) developments on the Prometheus side suggest Prometheus protobuf encoding should be favored.

See https://github.com/prometheus/OpenMetrics/issues/296#issuecomment-3875690035 and https://github.com/prometheus/OpenMetrics/pull/256#issuecomment-4090139930 for context.

I am open to contributing this feature, though would appreciate some guidance on whether we should replace OM proto or add an alternative encoder.

howardjohn:
Fixes https://github.com/prometheus/client_rust/issues/296

Replaces https://github.com/prometheus/client_rust/pull/294
Replaces https://github.com/prometheus/client_rust/pull/295

(I can split 294 and 295 out if desired just thought it was easier to handle all in one go)


CHANGELOG.md
```diff
@@ -4,6 +4,21 @@ All notable changes to this project will be documented in this file.
...
+
+- Added an off-by-default `protobuf-protox` feature to build protobuf support
+  without requiring an external `protoc` binary.
+
+### Changed
+
+- Updated `prost`, `prost-build`, and `prost-types` dependencies to `v0.14`.
+- The `protobuf` feature now generates and encodes Prometheus
+  `io.prometheus.client` protobuf messages from `metrics.proto` rather than the
+  OpenMetrics protobuf data model. This is a breaking change for users of the `protobuf` feature.
```
krisztianfekete, line 19: I think we should call out the most important change, namely the fact that metrics names will change.

krisztianfekete (approved):
I think all here is fine as the foundational work to enable Native Histograms implementation in follow-up PRs, and would love to see this move forward. Added one comment about the changelog.

brancz (approved):
lgtm

Looking forward to unlocking native histograms!

brancz:
(while I can merge, I'd feel more comfortable if @mxinden is the one who hits the merge button)

mxinden (commented):
Thanks for the context through https://github.com/prometheus/docs/pull/2836.

This library is explicitly build in a way to support many encoding formats. Instead of replacing the Open Metrics Protobuf format with the Prometheus Protobuf format, this pull request could as well add the Prometheus Protobuf format as a third encoding format, and mark the Open Metrics Protobuf format as `#[deprecated]`. In consecutive releases we can then remove the Open Metrics Protobuf format entirely. I assume this to be a much better user experience. @howardjohn what do you think? Would this be a large effort? Would this be worth the additional complexity.

As per protox, that looks like a good quality of life improvement.

I am sorry for the late review. I am no longer using the library myself and mostly focus on Firefox these days. Thus I think it would make sense for me to step down as a maintainer, handing it over to one or more of all of you fine folks. I will follow-up with a mail to the Prometheus mailing list.

howardjohn:
I'm good either way on adding or replacing. Adding seems to make sense, I just went with replacing based on https://github.com/prometheus/client_rust/issues/296#issuecomment-4161926959 (might have misinterpreted). I can switch it over to adding next week

brancz:
I don't know if I feel super strongly, since the whole point of all of this is to support native histograms, which won't be backward compatible with the current OM version supported by this library anyway.

krisztianfekete:
@mxinden, thank you for your effort on this project! Can you maybe cc. people here who should still be able to advise on directions going forward?

howardjohn (commented):
Ok i have sent another commit adding the legacy support back. Happy to keep it or remove it, just let me know

src/encoding.rs
```diff
@@ -13,17 +13,27 @@ use std::rc::Rc;
 use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
+#[cfg(feature = "legacy_protobuf")]
+#[cfg_attr(docsrs, doc(cfg(feature = "legacy_protobuf")))]
+pub mod openmetrics_protobuf;
```
krisztianfekete, line 26: Should we mark this deprecated as per https://github.com/prometheus/client_rust/pull/298#pullrequestreview-4175260290?

mxinden: Yes please.

src/encoding.rs
```diff
@@ -13,17 +13,27 @@ use std::rc::Rc;
 use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
+#[cfg(feature = "legacy_protobuf")]
```
mxinden: Preference for `openmetrics_protobuf`. The fact that it is legacy will be shown by the `#[deprecate]` tag.

src/encoding.rs
```diff
@@ -13,17 +13,27 @@ use std::rc::Rc;
...
+pub mod openmetrics_protobuf;
 #[cfg(feature = "protobuf")]
 #[cfg_attr(docsrs, doc(cfg(feature = "protobuf")))]
-pub mod protobuf;
+pub mod prometheus_protobuf;
 pub mod text;
 
+#[cfg(all(feature = "legacy_protobuf", not(feature = "protobuf")))]
+pub use openmetrics_protobuf as protobuf;
+#[cfg(feature = "protobuf")]
```
mxinden: Preference for `prometheus_protobuf`. One less breaking change, as `protobuf` was previously enabling the open metrics protobuf.

CHANGELOG.md
mxinden: I don't think the breaking change in itself is necessary here. Rust features are free. Let's just have two, i.e. `prometheus_protobuf` and `openmetrics_protobuf`.

mxinden (commented):
Thanks for the adjuments.

src/encoding.rs
howardjohn: Ok I pushed up:

protobuf OR openmetrics_protobuf => openmetrics, deprecated
prometheus_protobuf => prometheus

Sound right?

krisztianfekete:
Hey @mxinden, please let us know if we should change anything in the PR! Thanks!

mxinden (approved):
Thanks for avoiding the unnecessary breaking changes. Looks good to me.

I will let @brancz merge whenever he thinks this is ready :)

PR #299, opened 2026-04-24
feat(encoding): Expose `encode_descriptor` and `encode_metric`

patrick-ogrady:
For callers implementing their own `Registry` (https://github.com/commonwarexyz/monorepo/pull/3664), exposing access to more `encode_*` primitives (like `encode_eof`) avoids some hairy workarounds downstream.

### DIFF
--- a/src/encoding/text.rs
+++ b/src/encoding/text.rs
@@ -37,7 +37,9 @@
 //! assert_eq!(expected_msg, buffer);
 //! ```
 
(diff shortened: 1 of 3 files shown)


src/encoding/text.rs
```diff
@@ -182,6 +184,118 @@ where
...
+    M: EncodeMetric + ?Sized,
+{
+    metric.encode(
+        MetricEncoder {
+            writer,
+            prefix: None,
+            name,
+            unit,
+            const_labels,
+            family_labels: None,
```
patrick-ogrady, line 293: I think we could expose both `prefix`/`family_labels` here?

Issue #150, opened 2023-06-06
Implement native histograms


SuperQ:
Prometheus now supports a new "Native Histograms" datatype. We should implement this here.

mxinden:
:+1: from my end. I won't get to it any time soon. Contributions welcome.

nastynaz:
+1 would love to see this feature implemented

howardjohn:
This will be blocked on https://github.com/prometheus/client_rust/issues/296. I have some drafts for this and would like to contribute.

bengsparks:
Should this be closed due to https://github.com/prometheus/client_rust/pull/304 having been merged?

krisztianfekete:
Yes, thank you!

PR #304, opened 2026-05-28
Add native histogram support

howardjohn:
This PR adds support for native histograms using the new prometheus protobuf support recently merged.

The implementation is basically 1:1 lifted from the Golang implementation as much as possible, including tests.

Users can make any of the 3: classic only, native only, or classic + native histogram.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus-client"
-version = "0.25.0"
+version = "0.26.0"
 authors = ["Max Inden <mail@max-inden.de>"]
 edition = "2021"
 description = "Open Metrics client library allowing users to natively instrument applications."
@@ -65,6 +65,10 @@ harness = false
 
 [[bench]]
 name = "family"
+harness = false
+
+[[bench]]
+name = "histogram"
 harness = false
 
 [[bench]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus-client"
-version = "0.26.0"
+version = "0.25.0"
 authors = ["Max Inden <mail@max-inden.de>"]
 edition = "2021"
 description = "Open Metrics client library allowing users to natively instrument applications."
--- a/src/encoding/text.rs
+++ b/src/encoding/text.rs
@@ -37,7 +37,9 @@
 //! assert_eq!(expected_msg, buffer);
 //! ```
 
-use crate::encoding::{EncodeExemplarTime, EncodeExemplarValue, EncodeLabelSet, NoLabelSet};
+use crate::encoding::{
+    EncodeExemplarTime, EncodeExemplarValue, EncodeLabelSet, NativeHistogram, NoLabelSet,
+};
 use crate::metrics::exemplar::Exemplar;
 use crate::metrics::MetricType;
 use crate::registry::{Prefix, Registry, Unit};
@@ -444,6 +446,21 @@ impl MetricEncoder<'_> {
         Ok(())
     }
 
+    pub fn encode_histogram_with_native<S: EncodeLabelSet>(
+        &mut self,
+        sum: f64,
+        count: u64,
+        buckets: &[(f64, u64)],
+        exemplars: Option<&HashMap<usize, Exemplar<S, f64>>>,
+        _native: NativeHistogram<'_>,
+    ) -> Result<(), std::fmt::Error> {
+        if buckets.is_empty() {
+            return Err(std::fmt::Error);
+        }
+
+        self.encode_histogram(sum, count, buckets, exemplars)
+    }
+
     /// Encode an exemplar for the given metric.
     fn encode_exemplar<S: EncodeLabelSet, V: EncodeExemplarValue>(
         &mut self,
@@ -745,7 +762,7 @@ mod tests {
     use crate::metrics::exemplar::HistogramWithExemplars;
     use crate::metrics::family::Family;
     use crate::metrics::gauge::Gauge;
-    use crate::metrics::histogram::{exponential_buckets, Histogram};
+    use crate::metrics::histogram::{exponential_buckets, Histogram, NativeHistogramConfig};
     use crate::metrics::info::Info;
     use crate::metrics::{counter::Counter, exemplar::CounterWithExemplar};
     use pyo3::{prelude::*, types::PyModule};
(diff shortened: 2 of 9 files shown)


f41gh7:
Related issue https://github.com/prometheus/client_rust/issues/150

src/metrics/histogram.rs
```diff
@@ -173,15 +471,874 @@ pub fn linear_buckets(start: f64, width: f64, length: u16) -> impl Iterator<Item
...
+}
+
+fn enforce_bucket_limit(inner: &mut NativeHistogramState) -> bool {
+    if inner.max_buckets == 0 {
+        return false;
+    }
+
+    if inner.positive.len() + inner.negative.len() <= inner.max_buckets {
+        return false;
+    }
```
krisztianfekete, line 1180: Can we follow what client_golang is doing here:
https://github.com/prometheus/client_golang/blob/02eaf492695d380c231404841c5652a39b06be84/prometheus/histogram.go#L923-L962

basically do one widening/doubling per call then reassess as

> "very sparse buckets could lead to a low
  reduction of the bucket count (or even no reduction at all)"

src/metrics/histogram.rs
```diff
@@ -63,19 +303,55 @@ impl Histogram {
     /// let histogram = Histogram::new([10.0, 100.0, 1_000.0]);
     /// ```
     pub fn new(buckets: impl IntoIterator<Item = f64>) -> Self {
+        Self::new_classic(buckets)
+    }
+
+    /// Create a new classic [`Histogram`].
+    pub fn new_classic(buckets: impl IntoIterator<Item = f64>) -> Self {
```
krisztianfekete: Is this the same as `new`?

src/metrics/histogram.rs
```diff
@@ -173,15 +471,874 @@ pub fn linear_buckets(start: f64, width: f64, length: u16) -> impl Iterator<Item
...
+    if floor >= -(SCHEMA_MIN as f64) {
+        return SCHEMA_MIN;
+    }
+
+    -(floor as i8)
+}
+
+fn native_histogram_bounds() -> &'static [&'static [f64]; 9] {
+    &NATIVE_HISTOGRAM_BOUNDS
+}
```
krisztianfekete: Is there any reason why we cannot just use `NATIVE_HISTOGRAM_BOUNDS` directly?

src/metrics/histogram.rs
```diff
@@ -173,15 +471,874 @@ pub fn linear_buckets(start: f64, width: f64, length: u16) -> impl Iterator<Item
...
+    let mut index = if schema > 0 {
+        let bounds = &native_histogram_bounds()[schema as usize];
+        let bucket = bounds.partition_point(|bound| *bound < frac) as i32;
+        bucket + (exp - 1) * bounds.len() as i32
+    } else {
+        let mut key = exp;
+        if frac == 0.5 {
+            key -= 1;
+        }
+        let offset = (1_i32 << (-(schema as i32) as u32)) - 1;
```
krisztianfekete: We had https://github.com/prometheus/client_rust/pull/281 removing `as` casts, so maybe we should this pattern here and at a few other places? cc. @jalil-salame

src/encoding/text.rs
```diff
@@ -444,6 +446,21 @@ impl MetricEncoder<'_> {
...
+    pub fn encode_histogram_with_native<S: EncodeLabelSet>(
+        &mut self,
+        sum: f64,
+        count: u64,
+        buckets: &[(f64, u64)],
+        exemplars: Option<&HashMap<usize, Exemplar<S, f64>>>,
+        _native: NativeHistogram<'_>,
+    ) -> Result<(), std::fmt::Error> {
+        if buckets.is_empty() {
+            return Err(std::fmt::Error);
```
krisztianfekete, line 458: Can we have a typed error here and surface that the text encoder doesn't support native histograms?

src/metrics/histogram.rs
```diff
@@ -53,6 +65,234 @@ pub(crate) struct Inner {
...
+        Self::with_schema(pick_schema(bucket_factor))
+    }
+
+    /// Set a custom zero threshold.
+    pub fn zero_threshold(mut self, zero_threshold: f64) -> Self {
+        assert!(zero_threshold.is_finite());
+        assert!(zero_threshold >= 0.0 || zero_threshold == NATIVE_HISTOGRAM_ZERO_THRESHOLD_ZERO);
+        self.zero_threshold = zero_threshold;
+        self
+    }
```
krisztianfekete, line 137: Can we match the logic in client_go, where we have this: https://github.com/prometheus/client_golang/blob/02eaf492695d380c231404841c5652a39b06be84/prometheus/histogram.go#L576-L587

src/metrics/histogram.rs
```diff
@@ -173,15 +471,874 @@ pub fn linear_buckets(start: f64, width: f64, length: u16) -> impl Iterator<Item
 
 impl EncodeMetric for Histogram {
```
krisztianfekete, line 506: When using examplars this is skipped: https://github.com/prometheus/client_rust/blob/0c3ffbd92a9800343fb307626d970d23e2a4e1d1/src/metrics/exemplar.rs#L275-L285


and `histogram.get()` is classic only. 

So maybe we can move this into a helper and use that from `HistogramWithExemplars` as well?

krisztianfekete (commented):
This looks great, I am planning to do some actual testing as well soon, but added a couple of comments/questions in the meantime!

src/encoding/text.rs
howardjohn: Not really. the EncodeMetric itself has `std::fmt::Error` as the error type which conveys no information. To do so we would have to modify the public trait which doesn't seem in scope for this.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "prometheus-client"
-version = "0.25.0"
+version = "0.26.0"
```
krisztianfekete: Can you please revert this back to 0.25.0, and move the changelog entry to unreleased to follow the previous release conventions (e.g. https://github.com/prometheus/client_rust/commit/4a6d40a55443d5b18f5be311d246c03e56f417d6)?

krisztianfekete (commented):
Awesome work, all looks good to me! We are matching client_golang for all edge cases as far as I can see. We also have a nice test coverage (even more granular at some places), maybe outside of exemplars (we can do that as a follow-up). Also did some local testing, and can confirm the same.

Added one comment about doing the version bump separately, otherwise LGTM!

src/metrics/histogram.rs
```diff
@@ -273,4 +1428,765 @@ mod tests {
...
+        h.observe(4.0);
+        h.observe(-2.0);
+        h.observe(0.0);
+
+        let inner = h.inner.lock();
+        let native = inner.native.as_ref().unwrap();
+        assert_eq!(4, inner.count);
+        assert_eq!(1, native.zero_count);
+        assert_eq!(2, native.positive.len());
+        assert_eq!(3, native.negative.len() + native.positive.len());
```
krajorama: nit, this is just weird

```suggestion
        assert_eq!(1, native.negative.len());
```

src/metrics/histogram.rs
```diff
@@ -273,4 +1428,765 @@ mod tests {
...
+
+        let inner = h.inner.lock();
+        let native = inner.native.as_ref().unwrap();
+        assert_eq!(8, native.schema);
+        assert_eq!(1, native.zero_count);
+        assert_eq!(1, native.positive.len());
+    }
+
+    #[test]
+    fn native_histogram_schedules_reset_after_degradation() {
```
krajorama, line 1547: I'd add one more version of this test where the reset was scheduled  , but the reset is triggered by the next `observe()` , not `snapshot()`. As far as I see that code path isn't tested, only scheduled reset with `snapshot` and forced reset without schedule in `native_histogram_resets_when_limit_is_hit_after_min_reset_duration`.

src/metrics/histogram.rs
```diff
@@ -273,4 +1428,765 @@ mod tests {
...
+
+        let snapshot = h.snapshot().unwrap();
+        let native = snapshot.native.unwrap();
+
+        assert_eq!(1, snapshot.count);
+        assert_eq!(0.0, native.zero_threshold);
+        assert_eq!(0, native.zero_count);
+        assert_eq!(vec![(0, 0)], native.positive.spans);
+        assert!(native.positive.deltas.is_empty());
+        assert!(native.negative.spans.is_empty());
```
krajorama, line 2200: nit: might as well check that negative deltas is empty

src/metrics/histogram.rs
```diff
@@ -273,4 +1428,765 @@ mod tests {
...
+            bucket_index(0, f64::MIN_POSITIVE / 2f64.powi(52), false)
+        );
+        assert_eq!(1024, bucket_index(0, f64::MAX, false));
+        assert_eq!(1025, bucket_index(0, f64::INFINITY, true));
+        assert_eq!(262144, bucket_index(8, f64::MAX, false));
+        assert_eq!(262145, bucket_index(8, f64::INFINITY, true));
+    }
+
+    #[test]
+    fn native_histogram_bounds_match_client_golang_constants() {
```
krajorama: This is pretty selective and doesn't match fully against client_golang in reality.
Would be nicer to double check the values against recalculating the whole matrix, so check against the math. This is after all an optimization so we don't have to calculate `bounds[schema][i] = 2^(i / 2^schema) / 2` in production.

krajorama (commented):
This is looking good, I've checked parity with Claude and spent about 40m going through the code. Mostly interested in the tests where I have some comments.

krisztianfekete:
@howardjohn, can you please fix the failing tests?

krisztianfekete (approved):
LGTM!

PR #306, opened 2026-06-12
ci: check in Cargo.lock and build with --locked

krisztianfekete:
Initial approach: 
 - Pinning time to 0.3.47 due to this upstream regression in today's release: https://github.com/time-rs/time/issues/783

New approach:
 - Check in Cargo.lock and pass `--locked` in CI for reproducible builds, as suggested in review. This holds the time dep at a version that compiles, working around the regression above.


mxinden:
Given this being a dev-dependency, how about checking in a `Cargo.lock`? Despite this being a library, I consider this a best practice for a reproducible CI anyways.

krisztianfekete:
> Given this being a dev-dependency, how about checking in a `Cargo.lock`? Despite this being a library, I consider this a best practice for a reproducible CI anyways.

That has been my original idea, but thought it's not the pattern we'd like to follow here given it has never been committed to the repo, and CI lacked `--locked`. Happy to switch to that approach instead.

krisztianfekete:
@mxinden, can you please take a look when you have a chance?

krisztianfekete:
@mxinden sorry for the ping, but I'd like to fix the upstream `time` issue and complete cutting the release, so was wondering if you could review the PR?

krisztianfekete:
@mxinden, the upstream problem has been resolved, my release is out, but could use a review/approve when you have a chance. Thank you!

korniltsev-grafanista:
This change aligns with https://blog.rust-lang.org/2023/08/29/committing-lockfiles/

krisztianfekete:
@mxinden, gentle ping.

mxinden (approved):
👍

PR #307, opened 2026-06-22
fix(derive-encode): emit friendly compile errors instead of panicking

### LINKED ISSUE #269: compilation error msgs could be better
For struct like

```rust
struct Unnamed(String);
```

The derive macro `EncodeLabelSet` emits

```
error: proc-macro derive panicked
 --> tests/build/friendly-compilation-error-msg.rs:7:49
  |
7 |     #[derive(Debug, Clone, PartialEq, Eq, Hash, EncodeLabelSet)]
  |                                                 ^^^^^^^^^^^^^^
  |
  = help: message: Can not derive Encode for struct with unnamed fields.
```

I think `proc-macro derive panicked` is a little bit confuse, and the end developer needs to continue reading down the error message to find the cause.

An usual practice is emit [`compile_error!`](https://doc.rust-lang.org/std/macro.compile_error.html), which show the causes more directly.

lohitkolluri:
## Summary

Replace `panic!` calls in the `EncodeLabelSet` and `EncodeLabelValue` proc-macro derives with `compile_error!` so unsupported shapes surface a direct diagnostic that points to the `#[derive(...)]` line the user wrote, instead of the current `error: proc-macro derive panicked` message that points into the proc-macro internals.

## Full changes

- All 7 `panic!` sites in `derive-encode/src/lib.rs` now emit `compile_error!` tokens.
- Unknown `#[prometheus(...)]` attributes (which previously panicked inside a closure) are validated in a pre-pass so the derive can return a `compile_error!` token stream before generating the impl.
- New trybuild `compile_fail` test in `derive-encode/tests/build/friendly-compilation-error-msg.rs` covers unnamed struct, unit struct, enum, union, and unknown-attribute paths. The expected `.stderr` is committed alongside it.

## Testing

- `cargo test -p prometheus-client-derive-encode --test lib` passes deterministically (with and without `TRYBUILD=overwrite`).
- `cargo clippy -p prometheus-client-derive-encode --all-targets -- -D warnings` clean.
- `cargo fmt --check -p prometheus-client-derive-encode` clean.

## Notes

For valid inputs the derive behavior is unchanged. The new diagnostics are user-facing; no downstream tooling currently parses these derive errors, so this is not a breaking change.

## Issues

Closes #269

> Developed with Claude Code assistance.

### DIFF
--- a/derive-encode/src/lib.rs
+++ b/derive-encode/src/lib.rs
@@ -11,6 +11,12 @@ use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
 use syn::DeriveInput;
 
+fn error_spanned<T: quote::ToTokens>(tokens: &T, msg: &str) -> TokenStream {
+    syn::Error::new_spanned(tokens, msg)
+        .to_compile_error()
+        .into()
+}
+
 /// Derive `prometheus_client::encoding::EncodeLabelSet`.
 #[proc_macro_derive(EncodeLabelSet, attributes(prometheus))]
 pub fn derive_encode_label_set(input: TokenStream) -> TokenStream {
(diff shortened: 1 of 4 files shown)


derive-encode/src/lib.rs
```diff
@@ -17,6 +17,33 @@ pub fn derive_encode_label_set(input: TokenStream) -> TokenStream {
...
+                if !a.path().is_ident("prometheus") {
+                    continue;
+                }
+                match a.parse_args::<syn::Ident>() {
+                    Ok(ident) if ident == "flatten" => {}
+                    Ok(other) => {
+                        let msg = format!(
+                            "Provided attribute '{other}', but only 'flatten' is supported"
+                        );
+                        return quote! { compile_error!(#msg); }.into();
```
krisztianfekete: We could, instead of hand crafting the `compile_error!`, you could use `syn::Error::new_spanned(&other, msg).to_compile_error()`. It generates the same kind of error, but the little `^^^^` arrow points right at the unknown in `[prometheus(unknown)]` instead of  `#[derive(...)]` 

We could do these for the other sites too, `syn::Error::new_spanned(&ast, "...")` for the enum/union/unnamed-struct cases points at the offending type.

derive-encode/src/lib.rs
```diff
@@ -27,13 +54,7 @@ pub fn derive_encode_label_set(input: TokenStream) -> TokenStream {
                         .iter()
                         .find(|a| a.path().is_ident("prometheus"))
                         .map(|a| a.parse_args::<syn::Ident>().unwrap().to_string());
```
krisztianfekete: This `.unwrap()` is only safe because the pre-pass already filtered out unparseable attrs. This is easy to break later, and it means we parse every `#[prometheus(...)]` twice. Having the closure return a `Result` and `.collect()`-ing it would drop the pre-pass, the double parse, and the unwrap all at once.

krisztianfekete (commented):
Hey, thanks for the PR, left a few comments!

derive-encode/src/lib.rs
lohitkolluri: Switched all error sites to `syn::Error::new_spanned(...).to_compile_error()` so the span points at the offending token/type (e.g. the `unknown` ident inside `#[prometheus(unknown)]`, and the enum/union/unnamed-struct definition), instead of the derive line.

derive-encode/src/lib.rs
lohitkolluri: Dropped the pre-pass and the second parse; we now parse `#[prometheus(...)]` once during field handling and return a spanned compile error on invalid/unsupported args. That also removes the `unwrap()` and avoids parsing the attribute twice.

krisztianfekete (commented):
@lohitkolluri, can you please rebase this?

lohitkolluri:
@krisztianfekete Done!

krisztianfekete (approved):

krisztianfekete:
Thanks!

PR #309, opened 2026-07-03
ci: add github-actions ecosystem to dependabot

krisztianfekete:
Keep the existing cargo updates and add github-actions so Dependabot keeps maintained CI actions current. Avoids adopting PR #308, which would have dropped the cargo ecosystem and added Go-specific config irrelevant to this Rust crate.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"


krisztianfekete:
@SuperQ, would you mind reviewing this? You did review a similar one recently. Then I am planning to close #308. We should add this repo to the exclusion list for the bot I guess as well.

SuperQ (commented):
I would also add this to the top of the file to avoid the sync script from running on this file.

```
###
# Avoid syncing from prometheus/prometheus: no_prometheus_repo_sync
###
```

SuperQ (approved):

PR #324, opened 2026-08-03
(ci): replace archived actions-rs actions

krisztianfekete:
`actions-rs/toolchain` and `actions-rs/cargo` are both archived and still on `node12`, so dependabot keeps opening PRs to bump it, see #318.

This PR replaces them with `dtolnay/rust-toolchain`, which is already pinned at the same SHA in the cross-compile job, so no new dependency.
All six cargo commands are unchanged, `profile: minimal`/`override: true` are what dtolnay does by default, and `rustup component add` folds into `components:`.

Also dropped the protoc install from the Rustfmt job, since `cargo fmt` never builds anything.

Job names are untouched so the checks report exactly as before. Only downside: we losethe inline error annotations `actions-rs/cargo` gave us, but errors are still in the logs.

Closes #318.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,11 +12,9 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-      - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
+      - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
         with:
-          profile: minimal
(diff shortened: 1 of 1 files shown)


SuperQ:
The [GitHub Actions docs](https://docs.github.com/en/actions/tutorials/build-and-test-code/rust#specifying-a-rust-version) seem to indicate we could use `rustup` directly.

But this seems fine.

SuperQ (approved):

krisztianfekete:
> The [GitHub Actions docs](https://docs.github.com/en/actions/tutorials/build-and-test-code/rust#specifying-a-rust-version) seem to indicate we could use `rustup` directly.
> 
> But this seems fine.

Yeah, good point, that would work too. Went with `dtolnay/rust-toolchain` here mainly because it was already pinned in the cross-compile job, so it's not new surface, and we can avoid the extra `rustup component add` lines. We can consider migrating later, though. 

Thanks for the review!

PR #325, opened 2026-08-03
ci: avoid duplicate runs, harden checkout, use rust-cache

krisztianfekete:
`on: [push, pull_request]` ran every job twice for branches pushed to this repo, which is every dependabot PR, each `dependabot/*` branch has a 1 pair of `push` and `pull_request` runs per commit.

This PR:

- Scopes `push` to `master` and `v*`, and adds a concurrency group, so superseded pull request runs are cancelled while postsubmit runs on `master` (keyed by SHA) never cancel each other. No required status check depends on the `push` event, and the README badge (`?event=push`) still resolves, since pushes to `master` continue to fire.
- Sets `persist-credentials: false` on every checkout and declares `permissions: contents: read`. Neither is needed by any step here — nothing pushes, and `setup-protoc` receives its token through an explicit `repo-token:` input rather than from git config. This matters a little more than usual because `cargo test --all-features` executes `build.rs` from every dependency in the tree.
- Replaces `actions/cache` with `Swatinem/rust-cache`. The old key hashed `Cargo.toml` and not `Cargo.lock`, so lockfile-only dependabot PRs restored a cache built from a different dependency set, and `target/` was never cached at all. `save-if` limits saving to `master` so pull request branches cannot evict its cache. Pinned to `c1937114`, the `v2.9.1` release commit, per this repo's SHA-pin convention.

Fork PRs were already unaffected, their `push` fires in the fork, so the duplicate-run fix is not observable on this PR. It will show on the next dependabot PR.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,6 +2,8 @@ on:
   push:
     branches:
       - master
+    tags:
+      - "v*"
   pull_request:
 
 name: Continuous integration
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -1,7 +1,21 @@
-on: [push, pull_request]
+on:
+  push:
```
SuperQ, line 2: I'd also recommend tags.

```suggestion
  push:
    tags:
      - "v*"
```

.github/workflows/rust.yml
```diff
@@ -1,7 +1,21 @@
...
+    branches:
+      - master
+  pull_request:
 
 name: Continuous integration
 
+permissions:
+  contents: read
+
+concurrency:
```
SuperQ, line 14: It will be interesting to see how this works, I was thinking about doing this across other Prometheus repos.

.github/workflows/rust.yml
krisztianfekete: Thanks, added this!

ArthurSens (approved):
rubber stamping!

